PR #1, opened 2022-08-11
Basic Kotlin project structure

quietbits:
image


.gitignore
```diff
@@ -0,0 +1,15 @@
+*.iml
+.gradle
+/local.properties
+/.idea/caches
```
lijamie98, line 4: `.idea/` will ignore all files generated by IntelliJ or Android Studio.

build.gradle.kts
```diff
@@ -0,0 +1,31 @@
...
+    testImplementation(kotlin("test"))
+    implementation("com.github.stellar:java-stellar-sdk:0.34.2")
+}
+
+tasks.test {
+    useJUnitPlatform()
+}
+
+tasks.withType<KotlinCompile> {
+    kotlinOptions.jvmTarget = "1.8"
```
lijamie98, line 26: 1.8 is a little low. It should be safe to use 1.11 which is more modern and is one of the most popular version for backward compatibility.

build.gradle.kts
```diff
@@ -0,0 +1,31 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
```
lijamie98, line 1: I am not sure if we would want to use Kotlin DSL for Gradle. Although the anchor platform uses Kotlin DSL, most public Gradle examples are written in Groovy.

src/main/kotlin/org/stellar/kotlinwalletsdk/Wallet.kt
```diff
@@ -0,0 +1,5 @@
+package org.stellar.kotlinwalletsdk
```
lijamie98, line 1: In common practice, package names do not use `kotlin`. 

for example: `org.stellar.walletsdk` or `org.stellar.wallet.sdk`

lijamie98:
`.idea` folder should be in `.gitignore`

build.gradle.kts
quietbits: Sure, we can use Groovy for Gradle. I saw Kotlin option and just went all in on Kotlin. :sweat:

src/main/kotlin/org/stellar/kotlinwalletsdk/Wallet.kt
quietbits: If we later create Wallet SDK in Java, how would we name it? I added Kotlin there for that reason, but we can remove it.

quietbits:
@lijamie98, thanks for your review! I included `.idea` to "force" the same formatting for Kotlin files. In JavaScript, there is Prettier that does that. Is there a way to do the same in Kotlin/Java? Or maybe it's not a convention used in Kotlin/Java?

lijamie98 (approved):
LGTM

lijamie98:
> @lijamie98, thanks for your review! I included `.idea` to "force" the same formatting for Kotlin files. In JavaScript, there is Prettier that does that. Is there a way to do the same in Kotlin/Java? Or maybe it's not a convention used in Kotlin/Java?

Yes, we can enable Google Jave Style and ktfmt plug-in for Kotlin in Gradle. I can help with that. 

You may enable both plugins in IntelliJ that will help with automatic formatting

PR #2, opened 2022-08-15
Create account keys + ktfmt plugin

quietbits:
- `Wallet.create()` method to generate account keypair (doesn't depend on the network)
  - [x] Method
  - [x] Test
  - [x] Documentation
  - [x] Example (in `Main.kt`)
- Added `ktfmt` plugin for code formatting

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,19 @@
 # Kotlin Wallet SDK
 
 Kotlin Wallet SDK is a library that allows developers to build wallet applications on the Stellar network faster. It
-utilizes [Java Stellar SDK](https://github.com/stellar/java-stellar-sdk) to communicate with a Stellar Horizon network.
+utilizes [Java Stellar SDK](https://github.com/stellar/java-stellar-sdk) to communicate with a Stellar Horizon server.
 
 ## Wallet
 
-`class Wallet(horizonUrl: String, horizonPassphrase: String)`
+`class Wallet(horizonUrl: String, networkPassphrase: String)`
 
 Example
 
 ```kotlin
+import org.stellar.sdk.Network
+
 // Using Horizon testnet configuration
-val wallet = Wallet("https://horizon-testnet.stellar.org", "Test SDF Network ; September 2015")
+val wallet = Wallet("https://horizon-testnet.stellar.org", Network.TESTNET.toString())
 ```
 
 ### `Wallet.create()`
@@ -23,6 +25,6 @@ Example
 ```kotlin
 val newAccountKeypair = wallet.create()
 
-// publicKey=GB7S6OXH4UBA4QLR2OR4IXAOW7KOG67SUBEIXHGMJDNRKADHW7QGETPQ,
-// secretKey=SBTLU2GR5FYAQQ63HP65L3HF3M5PNKLX6E2TA2S4T63TZJHZDXE2GI27
+// Addresses are shortened for example only
+// publicKey=GB7S...GETPQ, secretKey=SBTL...GI27
 ```
--- a/README.md
+++ b/README.md
@@ -2,3 +2,27 @@
 
 Kotlin Wallet SDK is a library that allows developers to build wallet applications on the Stellar network faster. It
 utilizes [Java Stellar SDK](https://github.com/stellar/java-stellar-sdk) to communicate with a Stellar Horizon network.
+
+## Wallet
+
+`class Wallet(horizonUrl: String, horizonPassphrase: String)`
+
+Example
+
+```kotlin
+// Using Horizon testnet configuration
+val wallet = Wallet("https://horizon-testnet.stellar.org", "Test SDF Network ; September 2015")
+```
+
+### `Wallet.create()`
+
+`fun create(): AccountKeypair(val publicKey: String, val secretKey: String)`
+
+Example
+
+```kotlin
+val newAccountKeypair = wallet.create()
+
+// publicKey=GB7S6OXH4UBA4QLR2OR4IXAOW7KOG67SUBEIXHGMJDNRKADHW7QGETPQ,
+// secretKey=SBTLU2GR5FYAQQ63HP65L3HF3M5PNKLX6E2TA2S4T63TZJHZDXE2GI27
+```
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,7 @@
 plugins {
     id 'org.jetbrains.kotlin.jvm' version '1.7.10'
     id 'application'
+    id 'com.ncorti.ktfmt.gradle' version '0.8.0'
 }
 
 group = 'org.stellar'
@@ -30,4 +31,8 @@ compileTestKotlin {
 
 application {
     mainClassName = 'MainKt'
+}
+
+ktfmt {
+    googleStyle()
 }
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 plugins {
     id 'org.jetbrains.kotlin.jvm' version '1.7.10'
+    id 'com.diffplug.spotless' version '6.9.1'
     id 'application'
-    id 'com.ncorti.ktfmt.gradle' version '0.8.0'
 }
 
 group = 'org.stellar'
@@ -9,8 +9,21 @@ version = '1.0-SNAPSHOT'
 
 repositories {
     mavenCentral()
-    maven { url "https://jitpack.io" }
+    maven { url 'https://jitpack.io' }
 }
+
+
+spotless {
+    kotlin {
+        ktfmt("0.39").googleStyle()
+    }
+}
+
+afterEvaluate {
+    // Applying spotless fixes
+    tasks.getByName('spotlessKotlinCheck').dependsOn(tasks.getByName('spotlessKotlinApply'))
+}
+
 
 dependencies {
     testImplementation 'org.jetbrains.kotlin:kotlin-test'
@@ -31,8 +44,4 @@ compileTestKotlin {
 
 application {
     mainClassName = 'MainKt'
-}
-
-ktfmt {
-    googleStyle()
 }
--- a/src/main/kotlin/org/stellar/walletsdk/Wallet.kt
+++ b/src/main/kotlin/org/stellar/walletsdk/Wallet.kt
@@ -3,16 +3,12 @@ package org.stellar.walletsdk
 import org.stellar.sdk.KeyPair
 import org.stellar.sdk.Network
 import org.stellar.sdk.Server
-import shadow.com.google.gson.annotations.SerializedName
 
 class Wallet(private val horizonUrl: String, private val networkPassphrase: String) {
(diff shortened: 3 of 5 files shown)


build.gradle
```diff
@@ -30,4 +31,8 @@ compileTestKotlin {
 
 application {
     mainClassName = 'MainKt'
+}
+
+ktfmt {
+    googleStyle()
```
quietbits, line 37: Is there anything else we need to configure for this?

src/test/kotlin/org/stellar/walletsdk/WalletTest.kt
```diff
@@ -1,5 +1,25 @@
...
+    val publicKey = accountKeys.publicKey
+    val secretKey = accountKeys.secretKey
+
+    // Public key
+    assertEquals(56, publicKey.length)
+    assertTrue(publicKey.startsWith("G"))
+
+    // Secret key
+    assertEquals(56, secretKey.length)
+    assertTrue(secretKey.startsWith("S"))
```
quietbits, line 23: Doing a very basic sanity check that both keys are generated.

README.md
```diff
@@ -1,4 +1,28 @@
 # Kotlin Wallet SDK
 
 Kotlin Wallet SDK is a library that allows developers to build wallet applications on the Stellar network faster. It
-utilizes [Java Stellar SDK](https://github.com/stellar/java-stellar-sdk) to communicate with a Stellar Horizon network.
\ No newline at end of file
+utilizes [Java Stellar SDK](https://github.com/stellar/java-stellar-sdk) to communicate with a Stellar Horizon network.
```
JakeUrban, line 4:
```suggestion
utilizes [Java Stellar SDK](https://github.com/stellar/java-stellar-sdk) to communicate with a Stellar Horizon server.
```

README.md
```diff
@@ -1,4 +1,28 @@
...
+
+## Wallet
+
+`class Wallet(horizonUrl: String, horizonPassphrase: String)`
```
JakeUrban, line 8: Lets be consistent with existing tools and use `networkPassphrase`.
```suggestion
`class Wallet(horizonUrl: String, networkPassphrase: String)`
```

README.md
```diff
@@ -1,4 +1,28 @@
...
+
+Example
+
+```kotlin
+// Using Horizon testnet configuration
+val wallet = Wallet("https://horizon-testnet.stellar.org", "Test SDF Network ; September 2015")
```
JakeUrban, line 14: We could use the Java SDK's hard-coded values
```suggestion
import org.stellar.sdk.Network;

val wallet = Wallet("https://horizon-testnet.stellar.org", Network.TESTNET.toString())
```

README.md
```diff
@@ -1,4 +1,28 @@
...
+### `Wallet.create()`
+
+`fun create(): AccountKeypair(val publicKey: String, val secretKey: String)`
+
+Example
+
+```kotlin
+val newAccountKeypair = wallet.create()
+
+// publicKey=GB7S6OXH4UBA4QLR2OR4IXAOW7KOG67SUBEIXHGMJDNRKADHW7QGETPQ,
```
lijamie98, line 26: Although this is commented out and the keys are test keys, some code analyzer will complain about these keys. It is better to read these from external configurations.

README.md
```diff
@@ -1,4 +1,28 @@
...
+Example
+
+```kotlin
+// Using Horizon testnet configuration
+val wallet = Wallet("https://horizon-testnet.stellar.org", "Test SDF Network ; September 2015")
+```
+
+### `Wallet.create()`
+
+`fun create(): AccountKeypair(val publicKey: String, val secretKey: String)`
```
JakeUrban, line 19: The Java SDK's `Keypair.random()` does the same thing as this function.

Is the intention for the wallet sdk's functionality to overlap with the base Java SDK, or to be supplemental?

build.gradle
lijamie98: It is interesting this works for you. We have to use `spotless` plugin to make `ktfmt` work with Gradle. 
https://github.com/stellar/java-stellar-anchor-sdk/blob/main/build.gradle.kts#L61 

Does this reformat your code when you run `./gradlew build`?

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
```diff
@@ -1,5 +1,22 @@
...
+
 class Wallet(private val horizonUrl: String, private val horizonPassphrase: String) {
-    // TODO: Wallet code goes here
-}
\ No newline at end of file
+  private val server = Server(this.horizonUrl)
+  private val network = Network(this.horizonPassphrase)
+
+  data class AccountKeypair(
+    @SerializedName("publicKey") val publicKey: String,
```
lijamie98, line 13: Do you use `gson` or plan to serialize this class to JSON format? If not, I think we may not need `@SerializedName` annotation.

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
```diff
@@ -1,5 +1,22 @@
...
+    @SerializedName("secretKey") val secretKey: String
+  )
+
+  // Create account keys, generate new keypair
+  fun create(): AccountKeypair {
+    val keypair: KeyPair = KeyPair.random()
+    return AccountKeypair(keypair.accountId, String(keypair.secretSeed))
```
JakeUrban, line 20: The `AccountKeypair.publicKey` can be derived from `AccountKeypair.secretKey`, so we don't need to pass both to the constructor.

JakeUrban (commented):
I added some edits to the README and have some questions on the approach we'll be taking for the wallet SDK vs. the base SDK.

build.gradle
quietbits: @lijamie98 Yes, it seems to work. This is what I get when I run `./gradlew build` with the wrong formatting. Is that expected?

image

README.md
quietbits: These are examples of generated keypairs. I'll use shortened version instead for security.

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
quietbits: @lijamie98 What would you recommend? For web, we would definitely use JSON, but I'm not sure if that's the best format for Kotlin.

README.md
quietbits: Is our preferred way to use SDK defaults? I went with a simple string to avoid importing SDK just for that, but we can change it.

README.md
quietbits: The idea is to make it easier for anyone to use Kotlin Wallet SDK. `Wallet.create()` is just a wrapper for `Keypair.random()` but it will be more consistent with all the other SDK methods.

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
quietbits: Wouldn't we be adding extra step developers would need to do after `Wallet.create()` method is called? It feels more convenient to return both keys.

JakeUrban: Oh we should definitely make sure the `AccountKeypair` object provides devs with both, but what I'm saying is that we don't need to pass both keys to the constructor. We can pass `secretKey` to the object, and should be able to fetch `publicKey` & `secretKey` values from it.

JakeUrban: The keypair objects have methods `fromSecretSeed()` and and `fromPublicKey()`. Maybe we should have a similar pattern.

README.md
JakeUrban: I don't feel strongly about it, but those constants are there to help. I've found myself using them more often lately because I don't like having to search around to find the right string.

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
quietbits: @lijamie98 Is this the right way to define return type in this case (in `create()` method)? :pray:

build.gradle
lijamie98: IntelliJ has [ktfmt plugin](https://plugins.jetbrains.com/plugin/14912-ktfmt) that may help to work with the IDE better.

lijamie98: Looks like your Gradle setting only identifies the problem rather than fixing it. You will need to apply that.

In [AP (Anchor Platform)](https://github.com/stellar/java-stellar-anchor-sdk/blob/main/build.gradle.kts#L39), we use `spotless` plugin to apply the formatting. It can be a good reference for you.

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
lijamie98: Java has two popular libraries `gson` and `jackson` to convert an object to JSON string. In AP, we use `gson`. 

`@SerializedName` is only used when 
1. You need to send the object in JSON notation. Otherwise, you don't have to annotate it.
2. [When to serialize a field that has different name from the Java field name](https://www.tutorialspoint.com/what-to-use-serializedname-annotation-using-gson-in-java#:~:text=The%20%40SerializedName%20annotation%20can%20be,field%20with%20the%20provided%20name.). 

In your case, your Java field name equals to the JSON field name. So that you do not have to annotate it. :-)

README.md
```diff
@@ -1,4 +1,30 @@
...
+
+`class Wallet(horizonUrl: String, networkPassphrase: String)`
+
+Example
+
+```kotlin
+import org.stellar.sdk.Network
+
+// Using Horizon testnet configuration
+val wallet = Wallet("https://horizon-testnet.stellar.org", Network.TESTNET.toString())
```
lijamie98, line 16: You may use static imports to further simplify. 
```suggestion
val wallet = Wallet("https://horizon-testnet.stellar.org", TESTNET.toString())
```

IntelliJ has auto code complete for static imports. ([Stack Overflow](https://stackoverflow.com/questions/4387134/intellij-static-import-completion))

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
quietbits: Thanks for the detailed explanation, Jamie! We don't need JSON here, so I removed `@SerializedName`. :rocket:

build.gradle
quietbits: :tada: It's fixing formatting with `spotless` now. Thank you!

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
quietbits: Let's keep it as is for now. We can change it later if we feel strongly about it.

lijamie98 (approved):
LGTM

PR #3, opened 2022-08-17
Wallet.fund() method with helpers + tests

quietbits:
- `Wallet.fund()` method to active/fund account
  - [ ] Method
  - [ ] Test
  - [ ] Documentation
  - [ ] Example (in `Main.kt`)
- Added `utils` for helper functions
  - `Transaction.buildTransaction()` to simplify building transactions in other methods
  - `Operation.createSponsoredOperation()` a wrapper to create sponsored operation
- Added `mockk` for testing

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,11 @@ val wallet = Wallet("https://horizon-testnet.stellar.org", Network.TESTNET.toStr
 
 ### `Wallet.create()`
 
-`fun create(): AccountKeypair(val publicKey: String, val secretKey: String)`
+```kotlin
+fun create(): AccountKeypair(
+val publicKey: String,
+val secretKey: String)
+```
 
 Example
 
@@ -27,4 +31,48 @@ val newAccountKeypair = wallet.create()
 
 // Addresses are shortened for example only
 // publicKey=GB7S...GETPQ, secretKey=SBTL...GI27
+```
+
+### `Wallet.fund()`
+
+```kotlin
+fun fund(
+    sourceAddress: String,
+    destinationAddress: String,
+    startingBalance: String = "1",
+    sponsorAddress: String = ""
+): Transaction
+```
+
+Example
+
+```kotlin
+val sourceAddress = "GAMQTINWD3YPP3GLTQZ4M6FKCCSRGROQLIIRVECIFC6VEGL5F64CND22"
+val destinationAddress = "GDEIYYWIVK24CCQ3Y4QNGEIBEFABTTCFBRTVNQZ43VPOUNQARO7ZEKJY"
+
+val sponsoredFundTransaction =
+    wallet.fund(sourceAddress, destinationAddress, sponsorAddress = sourceAddress)
+```
+
+## Utils
+
+### `Operation.createSponsoredOperation()`
+
+```kotlin
+fun createSponsoredOperation(
+    sponsorAddress: String,
+    accountAddress: String,
+    operation: Operation
+): List<Operation>
+```
+
+### `Transaction.buildTransaction()`
+
+```kotlin
+fun buildTransaction(
+    sourceAddress: String,
+    server: Server,
+    network: Network,
+    operations: List<Operation>
+): Transaction
 ```
--- a/src/main/kotlin/Main.kt
+++ b/src/main/kotlin/Main.kt
@@ -8,4 +8,16 @@ fun main() {
   // Create account, generate keypairs
   //  val newAccountKeypair = wallet.create()
   //  println(newAccountKeypair)
+
+  val sourceAddress = "GAMQTINWD3YPP3GLTQZ4M6FKCCSRGROQLIIRVECIFC6VEGL5F64CND22"
+  val destinationAddress = "GDEIYYWIVK24CCQ3Y4QNGEIBEFABTTCFBRTVNQZ43VPOUNQARO7ZEKJY"
+
+  // Fund account
+  //  try {
+  //    val fundTxn = wallet.fund(sourceAddress, destinationAddress)
+  //    val fundSponsoredTxn =
+  //      wallet.fund(sourceAddress, destinationAddress, sponsorAddress = sourceAddress)
+  //  } catch (e: Error) {
+  //    println(e.toString())
+  //  }
 }
--- a/src/main/kotlin/org/stellar/walletsdk/Wallet.kt
+++ b/src/main/kotlin/org/stellar/walletsdk/Wallet.kt
@@ -1,8 +1,8 @@
 package org.stellar.walletsdk
 
-import org.stellar.sdk.KeyPair
-import org.stellar.sdk.Network
-import org.stellar.sdk.Server
+import org.stellar.sdk.*
+import org.stellar.walletsdk.utils.Operation.Companion.createSponsoredOperation
+import org.stellar.walletsdk.utils.Transaction.Companion.buildTransaction
 
 class Wallet(private val horizonUrl: String, private val networkPassphrase: String) {
   private val server = Server(this.horizonUrl)
@@ -14,5 +14,36 @@ class Wallet(private val horizonUrl: String, private val networkPassphrase: Stri
   fun create(): AccountKeypair {
     val keypair: KeyPair = KeyPair.random()
     return AccountKeypair(keypair.accountId, String(keypair.secretSeed))
+  }
+
+  // Fund (activate) account
+  // TODO: ??? do we need to include add trustline operation here?
+  fun fund(
+    sourceAddress: String,
+    destinationAddress: String,
+    startingBalance: String = "1",
+    sponsorAddress: String = ""
+  ): Transaction {
+    val isSponsored = sponsorAddress.isNotBlank()
+
+    if (!isSponsored && startingBalance.toInt() < 1) {
+      throw Error("Starting balance must be at least 1 XLM for non-sponsored accounts")
+    }
+
+    val startBalance = if (isSponsored) "0" else startingBalance
+
+    val createAccountOp: CreateAccountOperation =
+      CreateAccountOperation.Builder(destinationAddress, startBalance)
+        .setSourceAccount(sourceAddress)
+        .build()
+
+    val operations: List<Operation> =
+      if (isSponsored) {
+        createSponsoredOperation(sponsorAddress, destinationAddress, createAccountOp)
+      } else {
+        listOfNotNull(createAccountOp)
+      }
+
+    return buildTransaction(sourceAddress, server, network, operations)
   }
 }
--- a/src/main/kotlin/org/stellar/walletsdk/utils/Operation.kt
+++ b/src/main/kotlin/org/stellar/walletsdk/utils/Operation.kt
@@ -1,0 +1,26 @@
+package org.stellar.walletsdk.utils
+
+import org.stellar.sdk.BeginSponsoringFutureReservesOperation
+import org.stellar.sdk.EndSponsoringFutureReservesOperation
+import org.stellar.sdk.Operation
+
+class Operation {
+  companion object {
+    fun createSponsoredOperation(
+      sponsorAddress: String,
+      accountAddress: String,
+      operation: Operation
+    ): List<Operation> {
+      return listOfNotNull(
+        // Start reserve sponsoring
+        BeginSponsoringFutureReservesOperation.Builder(accountAddress)
+          .setSourceAccount(sponsorAddress)
+          .build(),
+        // Operation to sponsor
+        operation,
+        // End reserve sponsoring
+        EndSponsoringFutureReservesOperation(accountAddress)
+      )
+    }
+  }
+}
--- a/src/main/kotlin/org/stellar/walletsdk/utils/Transaction.kt
+++ b/src/main/kotlin/org/stellar/walletsdk/utils/Transaction.kt
@@ -1,0 +1,36 @@
+package org.stellar.walletsdk.utils
+
+import org.stellar.sdk.Network
+import org.stellar.sdk.Operation
+import org.stellar.sdk.Server
+import org.stellar.sdk.Transaction
+import org.stellar.sdk.responses.AccountResponse
+
+class Transaction {
+  companion object {
+    fun buildTransaction(
+      sourceAddress: String,
+      server: Server,
+      network: Network,
+      operations: List<Operation>
+    ): Transaction {
+      val sourceAccount: AccountResponse?
+
+      try {
+        sourceAccount = server.accounts().account(sourceAddress)
+      } catch (e: Error) {
+        throw Error("Source account was not found")
+      }
+
+      // TODO: add memo
+      // TODO: update max fee
+      // TODO: add time bounds
+
+      return Transaction.Builder(sourceAccount, network)
+        .addOperations(operations)
+        .setBaseFee(500)
+        .setTimeout(180)
+        .build()
+    }
+  }
+}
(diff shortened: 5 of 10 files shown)


src/main/kotlin/org/stellar/walletsdk/Wallet.kt
```diff
@@ -15,4 +15,35 @@ class Wallet(private val horizonUrl: String, private val networkPassphrase: Stri
     val keypair: KeyPair = KeyPair.random()
     return AccountKeypair(keypair.accountId, String(keypair.secretSeed))
   }
+
+  // Fund (activate) account
+  // TODO: ??? do we need to include add trustline operation here?
```
quietbits, line 20: Left this comment to make sure we don't forget about it.

src/main/kotlin/org/stellar/walletsdk/utils/Operation.kt
```diff
@@ -0,0 +1,26 @@
+package org.stellar.walletsdk.utils
```
quietbits, line 1: I'm not sure if this is a proper way to structure helper functions like these, but the idea was to use `Opeation` class for any helpers that are related to operations, and `Transaction` for transactions. Please let me know if there is a better way to handle these. :pray:

src/main/kotlin/org/stellar/walletsdk/utils/Transaction.kt
```diff
@@ -0,0 +1,36 @@
...
+
+      try {
+        sourceAccount = server.accounts().account(sourceAddress)
+      } catch (e: Error) {
+        throw Error("Source account was not found")
+      }
+
+      // TODO: add memo
+      // TODO: update max fee
+      // TODO: add time bounds
```
quietbits, line 27: I don't want to spend too much time on these now, but we will need to add them at some point.

src/test/kotlin/org/stellar/walletsdk/WalletTest.kt
```diff
@@ -1,25 +1,59 @@
...
+  private val wallet = Wallet(Constants.HORIZON_URL, Constants.NETWORK_PASSPHRASE)
 
-  @Test
-  fun `create() generates Stellar public and secret keys`() {
-    val accountKeys = wallet.create()
-    val publicKey = accountKeys.publicKey
-    val secretKey = accountKeys.secretKey
+  @Nested
+  @DisplayName("create")
+  inner class Create {
```
quietbits, line 15: I added nested tests with a display name to make it more clear in tests output which test comes from where.

image

src/test/kotlin/org/stellar/walletsdk/WalletTest.kt
```diff
@@ -1,25 +1,59 @@
...
-    assertTrue(secretKey.startsWith("S"))
+      // Secret key
+      assertEquals(56, secretKey.length)
+      assertTrue(secretKey.startsWith("S"))
+    }
+  }
+
+  @Nested
+  @DisplayName("fund")
+  inner class Fund {
```
quietbits, line 34: I'd like to add more tests that check the operations in transactions but I didn't find an easy way to do it right now. For example, it would be nice to check if the starting balance for a sponsored fund transaction is `0`. We'll spend more time on tests later.

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
```diff
@@ -15,4 +15,35 @@ class Wallet(private val horizonUrl: String, private val networkPassphrase: Stri
...
+  fun fund(
+    sourceAddress: String,
+    destinationAddress: String,
+    startingBalance: String = "1",
+    sponsorAddress: String = ""
+  ): Transaction {
+    val isSponsored = sponsorAddress.isNotBlank()
+
+    if (!isSponsored && startingBalance.toInt() < 1) {
+      throw Error("Starting balance must be at least 1 XLM for non-sponsored accounts")
```
lijamie98, line 30: It may be helpful to start defining our own exceptions so that the caller can handle the exceptions according to the type of exception that we throw.

README.md
```diff
@@ -27,4 +31,48 @@ val newAccountKeypair = wallet.create()
...
+val sourceAddress = "GAMQTINWD3YPP3GLTQZ4M6FKCCSRGROQLIIRVECIFC6VEGL5F64CND22"
+val destinationAddress = "GDEIYYWIVK24CCQ3Y4QNGEIBEFABTTCFBRTVNQZ43VPOUNQARO7ZEKJY"
+
+val sponsoredFundTransaction =
+    wallet.fund(sourceAddress, destinationAddress, sponsorAddress = sourceAddress)
+```
+
+## Utils
+
+### `Operation.createSponsoredOperation()`
```
lijamie98, line 59: Kotlin API (like Java API) docs are usually done in [KDoc](https://kotlinlang.org/docs/kotlin-doc.html#kdoc-syntax). If it is not necessary to expose these functions in README, it is easier to do it in KDoc.

src/main/kotlin/Main.kt
```diff
@@ -8,4 +8,16 @@ fun main() {
   // Create account, generate keypairs
   //  val newAccountKeypair = wallet.create()
   //  println(newAccountKeypair)
+
```
lijamie98, line 11: If we do not intend to run the SDK, we may remove the Main.kt. 

If we want something that can be run to test developing code, JUnit is our friend.

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
```diff
@@ -1,8 +1,8 @@
 package org.stellar.walletsdk
 
-import org.stellar.sdk.KeyPair
-import org.stellar.sdk.Network
-import org.stellar.sdk.Server
+import org.stellar.sdk.*
+import org.stellar.walletsdk.utils.Operation.Companion.createSponsoredOperation
+import org.stellar.walletsdk.utils.Transaction.Companion.buildTransaction
 
 class Wallet(private val horizonUrl: String, private val networkPassphrase: String) {
```
lijamie98, line 7: My gut feeling is the Wallet class is going to take a lot more parameters to construct. If that's the case, the [builder pattern ](https://www.baeldung.com/kotlin/builder-pattern) can help to reduce the complexity of object construction.

The [Lombok ](https://projectlombok.org/) library is popular to make the builder pattern less tedious. But there are discussions of deprecating using Lombok with Kotlin. I am not very sure about this. But just bringing this to your attention. 

The current implementation is fine with me.

lijamie98, line 7:
```suggestion
class Wallet(private val horizonUrl: String = "https://horizon-testnet.stellar.org", private val networkPassphrase: String = "Test SDF Network ; September 2015") {
```

src/main/kotlin/org/stellar/walletsdk/utils/Operation.kt
```diff
@@ -0,0 +1,26 @@
...
+          .setSourceAccount(sponsorAddress)
+          .build(),
+        // Operation to sponsor
+        operation,
+        // End reserve sponsoring
+        EndSponsoringFutureReservesOperation(accountAddress)
+      )
+    }
+  }
+}
```
lijamie98, line 26: In Kotlin, you may define functions in a `.kt` file. 

This is an [example](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/integration-tests/src/main/kotlin/org/stellar/anchor/platform/TestHelper.kt). 

```suggestion
    fun createSponsoredOperation(
      sponsorAddress: String,
      accountAddress: String,
      operation: Operation
    ): List<Operation> {
      return listOfNotNull(
        // Start reserve sponsoring
        BeginSponsoringFutureReservesOperation.Builder(accountAddress)
          .setSourceAccount(sponsorAddress)
          .build(),
        // Operation to sponsor
        operation,
        // End reserve sponsoring
        EndSponsoringFutureReservesOperation(accountAddress)
      )
    }

```

src/test/kotlin/org/stellar/walletsdk/Constants.kt
```diff
@@ -0,0 +1,17 @@
...
+    const val HORIZON_URL = "https://horizon-testnet.stellar.org"
+    val NETWORK_PASSPHRASE = Network.TESTNET.toString()
+
+    const val ADDRESS_ACTIVE = "GAMQTINWD3YPP3GLTQZ4M6FKCCSRGROQLIIRVECIFC6VEGL5F64CND22"
+    const val ADDRESS_INACTIVE = "GAW7QECBN2OI4LS4UUA3FO65Y2QPGQS3SPMTMRYK5ZX4IRZXXJPRWOBN"
+
+    val OP_CREATE_ACCOUNT: CreateAccountOperation =
+      CreateAccountOperation.Builder(ADDRESS_ACTIVE, "1").setSourceAccount(ADDRESS_INACTIVE).build()
+  }
+}
```
lijamie98, line 17: I think we do not need the Constants class. 

```suggestion

    const val HORIZON_URL = "https://horizon-testnet.stellar.org"
    val NETWORK_PASSPHRASE = Network.TESTNET.toString()

    const val ADDRESS_ACTIVE = "GAMQTINWD3YPP3GLTQZ4M6FKCCSRGROQLIIRVECIFC6VEGL5F64CND22"
    const val ADDRESS_INACTIVE = "GAW7QECBN2OI4LS4UUA3FO65Y2QPGQS3SPMTMRYK5ZX4IRZXXJPRWOBN"

    val OP_CREATE_ACCOUNT: CreateAccountOperation =
      CreateAccountOperation.Builder(ADDRESS_ACTIVE, "1").setSourceAccount(ADDRESS_INACTIVE).build()
```

src/test/kotlin/org/stellar/walletsdk/WalletTest.kt
```diff
@@ -1,25 +1,59 @@
...
+internal class WalletTest {
+  private val wallet = Wallet(Constants.HORIZON_URL, Constants.NETWORK_PASSPHRASE)
 
-  @Test
-  fun `create() generates Stellar public and secret keys`() {
-    val accountKeys = wallet.create()
-    val publicKey = accountKeys.publicKey
-    val secretKey = accountKeys.secretKey
+  @Nested
+  @DisplayName("create")
```
lijamie98, line 14: 👍 I learned something today.

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
```diff
@@ -15,4 +15,35 @@ class Wallet(private val horizonUrl: String, private val networkPassphrase: Stri
     val keypair: KeyPair = KeyPair.random()
     return AccountKeypair(keypair.accountId, String(keypair.secretSeed))
   }
+
+  // Fund (activate) account
+  // TODO: ??? do we need to include add trustline operation here?
+  fun fund(
```
lijamie98, line 21: The function only returns the transaction. Shouldn't we submit the transaction to Horizon?

README.md
```diff
@@ -27,4 +31,48 @@ val newAccountKeypair = wallet.create()
...
+
+### `Transaction.buildTransaction()`
+
+```kotlin
+fun buildTransaction(
+    sourceAddress: String,
+    server: Server,
+    network: Network,
+    operations: List<Operation>
+): Transaction
```
JakeUrban, line 77: I think we need more arguments here. How would developers specify a memo? What about their max base fee? timebound?

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
quietbits: Yes, the Wallet class most likely will have more parameters, but it's hard to tell right now what exactly they will be. Let's optimize when we have more parameters. I'll keep in mind the builder pattern. :pray:

src/main/kotlin/org/stellar/walletsdk/utils/Operation.kt
quietbits: I'm so glad to see that not everything needs to go in class in Kotlin. :sweat_smile: I'll move helper methods directly to the `utils` folder. :+1:

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
quietbits: There will be separate functions to sign and submit. That way developers will have better control of how many operations they want to add, how they want to sign, etc.

README.md
```diff
@@ -27,4 +31,48 @@ val newAccountKeypair = wallet.create()
...
+```
+
+Example
+
+```kotlin
+val sourceAddress = "GAMQTINWD3YPP3GLTQZ4M6FKCCSRGROQLIIRVECIFC6VEGL5F64CND22"
+val destinationAddress = "GDEIYYWIVK24CCQ3Y4QNGEIBEFABTTCFBRTVNQZ43VPOUNQARO7ZEKJY"
+
+val sponsoredFundTransaction =
+    wallet.fund(sourceAddress, destinationAddress, sponsorAddress = sourceAddress)
```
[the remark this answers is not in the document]
quietbits: No, by default `sponsorAddress` is empty (we're checking if the value is not empty to see if sponsoring is enabled). This example might be a bit confusing without any context. We'll make sure to make it more clear in the real docs.

It's possible to have a different sponsor account than the source account, right?

README.md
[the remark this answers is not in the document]
quietbits: Good point. :+1: I renamed it to `sponsorOperation` since it was moved to its own file (outside of the class).

README.md
quietbits: Yup! I added `TODO:` comments in `buildTransaction` helper to add these later. Right now I'm focusing on the easy path, just to see how everything comes together.

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
quietbits: :100: I'll add a note to do that once we have more exceptions to consider.

README.md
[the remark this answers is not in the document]
JakeUrban: Yea I think the example is confusing, maybe we define a third public key and use that for the sponsor, but I don't feel strongly.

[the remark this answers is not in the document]
quietbits: I'll update the example :+1:

src/main/kotlin/org/stellar/walletsdk/utils/sponsorOperation.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import org.stellar.sdk.BeginSponsoringFutureReservesOperation
+import org.stellar.sdk.EndSponsoringFutureReservesOperation
+import org.stellar.sdk.Operation
+
+fun sponsorOperation(
+  sponsorAddress: String,
+  accountAddress: String,
+  operation: Operation
+): List<Operation> {
```
JakeUrban, line 11: I know you're keeping it simple right now, but it might be valuable to have some validation of the operations that are passed to make sure they are sponsor-able ops.

JakeUrban (approved):

src/main/kotlin/org/stellar/walletsdk/utils/sponsorOperation.kt
quietbits: :100: Thanks for catching this! I added a check in [this commit.](https://github.com/stellar/kotlin-wallet-sdk/pull/3/commits/62321063c6ecaad290e70b46bfc8ab8875349f1a)

PR #4, opened 2022-08-19
Wallet: add/remove asset support and account signer

quietbits:
- `Wallet.addAssetSupport()` and `Wallet.removeAssetSupport()` methods
  - [x] Method
  - [x] Test
  - [x] Documentation
- `Wallet.addAccountSigner()` and `Wallet.removeAccountSigner()` methods
  - [x] Method
  - [x] Test
  - [x] Documentation


README.md
```diff
@@ -57,6 +57,84 @@ val sponsoredFundTransaction =
...
+
+```kotlin
+fun addAssetSupport(
+    sourceAddress: String,
+    assetCode: String,
+    assetIssuer: String,
+    trustLimit: String = "",
+    sponsorAddress: String = ""
+): Transaction
+```
```
JakeUrban, line 70: I would rename this to `addAsset()`, no need to use the term support.

Also, how are we thinking about liquidity pools? From a horizon perspective, you need a trustline to a liquidity pool just like you need a trustline to an asset. But the parameters are a bit different, so are we planning on adding a function for this like `addLiquidityPool()`?

README.md
```diff
@@ -57,6 +57,84 @@ val sponsoredFundTransaction =
...
+
+### `Wallet.removeAssetSupport()`
+
+```kotlin
+fun removeAssetSupport(
+    sourceAddress: String,
+    assetCode: String,
+    assetIssuer: String
+): Transaction
+```
```
JakeUrban, line 89: Something I just thought about is why do we need to specify `sourceAddress` for all these functions? Can't we just use the `Wallet` object's `publicKey`? We could make `sourceAddress` optional for most functions and it would only be necessary if you wanted a different source address that the wallet's public key.

README.md
quietbits: `addAssetSupport` is the name we have in the document now. We can change it, but we'll need to check with Bruno about it.

We haven't thought about liquidity pools yet, but most likely it will be a separate function. :+1:

README.md
quietbits: I agree it's quite annoying to keep providing `sourceAddress`, but the `Wallet` doesn't have a mechanism to store the public key at the moment. We'll definitely need to make this better once we have an idea of how the public key will be accessed.

The `sourceAddress` does make sponsored transactions more clear, though.

JakeUrban: > the Wallet doesn't have a mechanism to store the public key at the moment

🤔 maybe we should think about this.

What if instead of having `Wallet.create()` return an `AccountKeypair`, the `Wallet` object stored a regular stellar SDK `Keypair`?

So creating a wallet could look like this:
```
Wallet(keypair=Keypair.fromSecret(<secret key>))
```

Or, if you're looking to create a new account:

```
Wallet()  // automatically creates a keypair using `Keypair.random()`
```

Then the `Wallet` class could expose `Wallet.publicKey` and `Wallet.secretKey`.

quietbits: :thinking: What if we create a new method `Wallet.setAccountAddress(publicKey: String)` (simple setter) that has to be called after `Wallet(...)` init. That way the logic could be like this:
1. For new sign-ups: backend has no public key -> Wallet() -> Wallet.create() -> save public key to backend -> Wallet.setAccountAddress(publicKey) -> ...
2. For sign-ins: get public key from backend -> Wallet() -> Wallet.setAccountAddress(publicKey) -> ...

I'd like to keep `Wallet.create()` as is because it would be a single place where keys are generated and a secret key is stored in Keystore right after. Nothing else in the code would "know" about it.

@lijamie98 what are your thoughts on this?

JakeUrban: I don't think it makes sense to have to use `Wallet.create()` and then `Wallet.setAccountAddress()` to save the public key to the same object that just generated it.

I think it would make more sense if there was no `Wallet.create()`, since its effectively equivalent to `Keypair.random()`. So it would look something like this:

```
kp = Keypair.random()
wallet = Wallet(accountAddress=kp.publicKey)
```
or
```
kp = Keypair.random()
wallet = Wallet()
wallet.setAccountAddress(kp.publicKey)
```

quietbits: @brunomuler, do we want to rethink what the` Wallet.create()` method does? If it just generates a keypair, maybe it could be moved to a helper method?

JakeUrban: This discussion doesn't have to block the PR, but lets keep talking about it. Personally I don't think it makes sense to create functions like `Wallet.create()`, which is already supported by our SDKs via `Keypair.random()`.

lijamie98:
If the sourceAccount will be discussed in the next PR, there is nothing blocking to merge the PR.

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
```diff
@@ -49,4 +49,66 @@ class Wallet(
 
     return buildTransaction(sourceAddress, server, network, operations)
   }
+
+  // Add trustline
+  fun addAssetSupport(
+    sourceAddress: String,
+    assetCode: String,
+    assetIssuer: String,
+    trustLimit: String = "",
```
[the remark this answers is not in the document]
quietbits: It's more convenient to have a default value there, otherwise, it's required when `sponsorAddress` is provided (it makes sense to keep this after `trustLimit`). Maybe there is a trick in Kotlin to make that happen?

Your comment made me catch a bug for the default value, which cannot be an empty string. :sweat_smile:  We're now setting it to `Long.MAX_VALUE` (Int64). I also added a few more tests to catch these types of errors sooner. :crossed_fingers:

PR #5, opened 2022-08-25
Wallet: validate and submit transaction

quietbits:
- Added utils:
  - `accountAvailableNativeBalance()`
  - `accountReservedBalance()`
  - `buildFeeBumpTransaction()`
  - `stroopsToLumens()` and `lumensToStroops()`
  - `validateTransaction()`
- Added `wallet.submitTransaction()`
- Added tests and updated readme

There is nothing to add for signing the transaction as it depends on KeyStore which is only available on the client (Android). We can add any helpers later if needed.


src/main/kotlin/org/stellar/walletsdk/Wallet.kt
```diff
@@ -111,4 +112,13 @@ class Wallet(
   fun removeAccountSigner(sourceAddress: String, signerAddress: String): Transaction {
     return addAccountSigner(sourceAddress, signerAddress, 0)
   }
+
+  // Submit transaction
+  fun submitTransaction(signedTransaction: Transaction): SubmitTransactionResponse {
```
quietbits, line 117: I'm not sure how valuable this method is. The only thing it does is provides a `server` internally.

There are no tests for this because I couldn't get `server` mock working through `wallet`. It worked when I used mocked `server` directly in the test. We probably don't need a test for this, anyway. :woman_shrugging:

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
```diff
@@ -111,4 +112,13 @@ class Wallet(
...
+    try {
+      return server.submitTransaction(signedTransaction)
+    } catch (e: Error) {
+      throw Error("Transaction failed: $e")
+    }
+  }
```
JakeUrban, line 123: I don't think this function is useful as-is, but I think we can make it useful by providing some default error handling.

For example, we could add retry logic that resubmits the transaction when a `504` or `insufficient_fee` is received.

Essentially, we could build it to provide two possible states: success (included in a ledger) or failure (transaction expired or transaction invalid).

src/main/kotlin/org/stellar/walletsdk/utils/accountAvailableNativeBalance.kt
```diff
@@ -0,0 +1,14 @@
...
+import org.stellar.sdk.responses.AccountResponse
+
+fun accountAvailableNativeBalance(account: AccountResponse): String {
+  val nativeAmount = account.balances.find { it.assetType == "native" }!!.balance.toBigDecimal()
+  val reservedBalance = accountReservedBalance(account).toBigDecimal()
+  val availableAmount = nativeAmount.minus(reservedBalance)
+
+  return if (availableAmount <= BigDecimal(0)) {
+    return "0"
+  } else availableAmount.toPlainString()
```
JakeUrban, line 13: `availableAmount` will never be less than 0 because the amount reserved can at most be the total amount held.

src/main/kotlin/org/stellar/walletsdk/utils/accountReservedBalance.kt
```diff
@@ -0,0 +1,23 @@
...
+  val buyingLiabilities =
+    account.balances.find { it.assetType == "native" }?.buyingLiabilities?.get() ?: "0"
+
+  return BigDecimal(BASE_RESERVE_MIN_COUNT)
+    .plus(subEntryCount)
+    .plus(numSponsoring)
+    .minus(numSponsored)
+    .times(BASE_RESERVE.toBigDecimal())
+    .plus(buyingLiabilities.toBigDecimal())
+    .toPlainString()
```
JakeUrban, line 22: The docs say this is the formula:

```
(2 + numSubEntries + numSponsoring - numSponsored) * baseReserve + liabilities.selling
```

I think you mixed up buying liabilities with selling liabilities

src/main/kotlin/org/stellar/walletsdk/utils/buildFeeBumpTransaction.kt
```diff
@@ -0,0 +1,23 @@
+package org.stellar.walletsdk.utils
+
+import org.stellar.sdk.FeeBumpTransaction
+import org.stellar.sdk.Server
+import org.stellar.sdk.Transaction
+
+fun buildFeeBumpTransaction(
+  feeAccount: String,
+  innerTransaction: Transaction,
+  maxFeeInStroops: Long,
```
JakeUrban, line 10: This is not a great name because its not the maximum fee, its the maximum _base_ fee. The transaction could be `# of ops` * `maxFeeInStroops`.

src/main/kotlin/org/stellar/walletsdk/utils/validateTransaction.kt
```diff
@@ -0,0 +1,24 @@
+package org.stellar.walletsdk.utils
+
+import org.stellar.sdk.Server
+import org.stellar.sdk.Transaction
+
+fun validateTransaction(transaction: Transaction, server: Server) {
```
JakeUrban, line 6: This name is too generic. I don't know what it does by reading the name, can we rename it to be more explicit?

JakeUrban (changes requested):

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
quietbits: We could retry when the response is `504` (timeout), but I'm not sure we should if the fee is too low. What if the transaction has a much lower fee than recommended? Or is it our recommendation for wallets to retry in both cases?

src/main/kotlin/org/stellar/walletsdk/utils/accountAvailableNativeBalance.kt
quietbits: The check is there just for safety. I don't think it hurts to have it there. If you feel strongly against it, we can remove it.

src/main/kotlin/org/stellar/walletsdk/utils/accountReservedBalance.kt
quietbits: My bad, thanks for catching this!

src/main/kotlin/org/stellar/walletsdk/utils/buildFeeBumpTransaction.kt
quietbits: :+1: Does `maxBaseFeeInStroops` sound good?

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
JakeUrban: > I'm not sure we should if the fee is too low. What if the transaction has a much lower fee than recommended?

If we don't retry in this case, then we'd need to throw an exception and let the developer decide what to do.

They have a couple options,

1. Call `submitTransaction()` again with the same transaction
2. Create a new transaction with a higher base fee and call `submitTransaction()` with the new transaction
3. Wrap the transaction in a `FeeBumpTransaction` and call `submitTransaction()` with the fee bump transaction
4. Error out in their app and inform the user that their transaction failed

I'm ok throwing an exception and letting them decide what to do, but we need to make sure we document how to handle the exception. We also should have helper functions that can fetch a base fee to be used from Horizon's `GET /fee_stats`.

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
JakeUrban: responded in another comment.

src/main/kotlin/org/stellar/walletsdk/utils/validateTransaction.kt
quietbits: So the idea was that this helper could be used to validate the transaction before submitting it. Then we could validate signatures, etc. We could also check it before signing. `preSubmitValidation()`? I'm open to suggestions. :pray:

JakeUrban: Hmm.. I don't know if we should have this function, because there are a lot of checks we could make, and these checks could be expensive.

See the errors defined in this section of the [docs](https://developers.stellar.org/api/errors/http-status-codes/horizon-specific/). I'd be concerned we're giving the developer a false sense of security because we're only checking subset of checks we could perform. Thoughts?

quietbits: Yeah, I'm on the fence about it. :thinking: I thought it might be useful to check if the account has enough balance to cover the fees, but maybe that's not necessary for a custodial wallet? I'm OK removing this helper.

src/main/kotlin/org/stellar/walletsdk/utils/buildTransaction.kt
```diff
@@ -16,7 +17,7 @@ fun buildTransaction(
 
   try {
     sourceAccount = server.accounts().account(sourceAddress)
-  } catch (e: Error) {
+  } catch (e: Exception) {
     throw Error("Source account was not found")
```
lijamie98, line 21: This may not be accurate because the `Exception` is an `IOException` which can happen while making HTTP connections to the Horizon server.

src/main/kotlin/org/stellar/walletsdk/utils/validateTransaction.kt
```diff
@@ -0,0 +1,24 @@
...
+  val sourceAccount =
+    try {
+      server.accounts().account(transaction.sourceAccount)
+    } catch (e: Exception) {
```
lijamie98, line 10: Same as above.

src/main/kotlin/org/stellar/walletsdk/utils/validateTransaction.kt
```diff
@@ -0,0 +1,24 @@
...
+      throw Error("Source account ${transaction.sourceAccount} does not exist")
+    }
+
+  val sourceAccountBalance = accountAvailableNativeBalance(sourceAccount).toBigDecimal()
+  val transactionFees = stroopsToLumens(transaction.fee.toString()).toBigDecimal()
+
+  if (sourceAccountBalance < transactionFees) {
+    throw Error(
```
lijamie98, line 18: Error is very generic. It is good to start defining Wallet SDK exceptions. Such as `InvalidArgumentException` so that the caller of the SDK can process the specific error and report to the customer. 

[`Error`](https://docs.oracle.com/javase/6/docs/api/java/lang/Error.html) is used to indicate serious problems that a reasonable caller should not try to catch. While `Exception` is the bread and butter for exception handling.

src/test/kotlin/org/stellar/walletsdk/helpers/objectFromJsonFile.kt
```diff
@@ -0,0 +1,15 @@
...
+
+fun <T> objectFromJsonFile(filePath: String, t: Class<T>): T {
+  val jsonString =
+    try {
+      File(filePath).readText()
+    } catch (e: Exception) {
+      throw Error("File not found")
+    }
+
+  return Gson().fromJson(jsonString, t)
```
lijamie98, line 14: This creates a Gson() object every time we call.  The anchor SDK uses the[ singleton pattern t](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/core/src/main/java/org/stellar/anchor/util/GsonUtils.java#L19)o avoid the problem and provides a single point of access if we want to[ inject adapters](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/core/src/main/java/org/stellar/anchor/util/GsonUtils.java#L15) not supported by Gson.

src/test/kotlin/org/stellar/walletsdk/helpers/objectFromJsonFile.kt
```diff
@@ -0,0 +1,15 @@
+package org.stellar.walletsdk.helpers
+
+import com.google.gson.Gson
+import java.io.File
+
+fun <T> objectFromJsonFile(filePath: String, t: Class<T>): T {
+  val jsonString =
+    try {
+      File(filePath).readText()
+    } catch (e: Exception) {
```
lijamie98, line 10: There are many exceptions that may happen while reading a file. Throwing `Error("File not found")` can be mis-leading later.

You may leave the exception un-caught and let the caller handle the Exception.

src/test/kotlin/org/stellar/walletsdk/helpers/objectFromJsonFile.kt
```diff
@@ -0,0 +1,15 @@
+package org.stellar.walletsdk.helpers
+
+import com.google.gson.Gson
+import java.io.File
+
+fun <T> objectFromJsonFile(filePath: String, t: Class<T>): T {
+  val jsonString =
+    try {
+      File(filePath).readText()
```
lijamie98, line 9: It is safer to use [try-with-resources](https://www.baeldung.com/kotlin/try-with-resources) to avoid resource leak because File implements auto-closable.

src/test/kotlin/org/stellar/walletsdk/utils/AccountReservedBalanceTest.kt
```diff
@@ -0,0 +1,28 @@
...
+      objectFromJsonFile("src/test/resources/account_basic.json", AccountResponse::class.java)
+    val reservedBalance = accountReservedBalance(account)
+
+    assertEquals(1.0, reservedBalance.toDouble())
+  }
+
+  @Test
+  fun `more complex account`() {
+    val account =
+      objectFromJsonFile("src/test/resources/account_full.json", AccountResponse::class.java)
```
lijamie98, line 23: The path `src/test/resources/account_full.json` is reliable because it depends on the running context of the JUnit test process. 

Here is a good example of [how to read a file from the resource folder](https://mkyong.com/java/java-read-a-file-from-resources-folder/).

src/main/kotlin/org/stellar/walletsdk/utils/validateTransaction.kt
quietbits: Thanks for the explanation, Jamie! I'll replace all `Error` with `Exception`.

src/main/kotlin/org/stellar/walletsdk/utils/validateTransaction.kt
quietbits: I'll keep this for now, but we can always remove it later.

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
```diff
@@ -111,4 +112,37 @@ class Wallet(
   fun removeAccountSigner(sourceAddress: String, signerAddress: String): Transaction {
     return addAccountSigner(sourceAddress, signerAddress, 0)
   }
+
+  // Submit transaction
+  fun submitTransaction(signedTransaction: Transaction, serverInstance: Server = server): Boolean {
+    val maxRetries = 3
```
lijamie98, line 118: Retries are sometimes necessary evil. We should probably let the caller decide how to handle retries.

If it is necessary to implement retries, we should parameterize the retries count and cool-down period.

src/main/kotlin/org/stellar/walletsdk/utils/buildFeeBumpTransaction.kt
```diff
@@ -0,0 +1,24 @@
...
+  feeAccount: String,
+  innerTransaction: Transaction,
+  maxBaseFeeInStroops: Long,
+  server: Server
+): FeeBumpTransaction {
+  try {
+    server.accounts().account(feeAccount)
+  } catch (e: IOException) {
+    throw Exception("Fee account was not found")
+  }
```
lijamie98, line 18: I remember this repeats a couple of times. It may make sense to have a function such as `validateAddress()` in the project.

src/main/kotlin/org/stellar/walletsdk/utils/accountAvailableNativeBalance.kt
```diff
@@ -0,0 +1,14 @@
+package org.stellar.walletsdk.utils
```
lijamie98, line 1: The Kotlin [convention of source files ](https://kotlinlang.org/docs/coding-conventions.html#source-file-names)starts with capital letter.

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
quietbits: :+1: I'll remove retries for now. We can decide later how we'd like to handle it. @JakeUrban

src/main/kotlin/org/stellar/walletsdk/utils/accountAvailableNativeBalance.kt
quietbits: @lijamie98, do you mean we should rename `utils` to `Utils` or all the file names should start with a capital letter? `utils` is a package inside `org.stellar.walletsdk`, so I used a lowercase name there. All the helper methods that don't have a class or interface are lowercase as well. I'm still learning Kotlin conventions, so I'm happy to change anything that is off. :sweat_smile:

lijamie98: Oh, what I meant is the file name `accountAvailableNativeBalance.kt` should be renamed to `AccountAvailableNativeBalance.kt`. 

When we rename the file, the name is different from the function. That is because Kotlin/Java developers tends to aggregate a couple of functions in one file.

For example, you may create the file `AccountHelper.kt` and group these functions in the same file. 
- accountAvailableNativeBalance
- accountReservedBalance
- fetchAccountFromAddress

lijamie98 (approved):
LGTM.

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
```diff
@@ -111,4 +111,25 @@ class Wallet(
   fun removeAccountSigner(sourceAddress: String, signerAddress: String): Transaction {
     return addAccountSigner(sourceAddress, signerAddress, 0)
   }
+
+  // Submit transaction
+  fun submitTransaction(signedTransaction: Transaction, serverInstance: Server = server): Boolean {
```
JakeUrban, line 116: can we change the name of `serverInstance` to `server`, and change the object's `server` object to `defaultServer`?

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
```diff
@@ -111,4 +111,25 @@ class Wallet(
...
+        return true
+      }
+
+      var errorMessage = "Transaction failed"
+
+      if (!response.extras?.resultCodes?.transactionResultCode.isNullOrBlank()) {
+        errorMessage += ": ${response.extras.resultCodes.transactionResultCode}"
+      }
+
+      throw Exception(errorMessage)
```
JakeUrban, line 130: Let's take this out of the try/catch clause, since its not the code that may throw an exception.

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
```diff
@@ -111,4 +111,25 @@ class Wallet(
...
+    } catch (e: Throwable) {
+      throw e
```
JakeUrban, line 132: We shouldn't have a try/catch at all if we just want to re-throw it.

If we decide we do want to do something about the error then we can keep the try/catch, but I think we need something more specific than `Throwable`. We should be catching specific errors raised from `submitTransaction()`.

src/main/kotlin/org/stellar/walletsdk/util/FetchAccountFromAddress.kt
```diff
@@ -0,0 +1,13 @@
+package org.stellar.walletsdk.util
+
+import java.io.IOException
+import org.stellar.sdk.Server
+import org.stellar.sdk.responses.AccountResponse
+
+fun fetchAccountFromAddress(accountAddress: String, server: Server): AccountResponse {
+  try {
+    return server.accounts().account(accountAddress)
+  } catch (e: IOException) {
```
JakeUrban, line 10: Isn't there a `NotFound` exception or something similar? `IOException` will raise an error if it cannot connect to horizon, but currently the SDK will inform the account wasn't found, which is inaccurate.

src/main/kotlin/org/stellar/walletsdk/util/FetchAccountFromAddress.kt
```diff
@@ -0,0 +1,13 @@
+package org.stellar.walletsdk.util
+
+import java.io.IOException
+import org.stellar.sdk.Server
+import org.stellar.sdk.responses.AccountResponse
+
+fun fetchAccountFromAddress(accountAddress: String, server: Server): AccountResponse {
```
JakeUrban, line 7: Why not `fetchAccount()` or `fetchAccountData()`? Its obvious thats if from the address because it requires that parameter.

src/main/kotlin/org/stellar/walletsdk/util/ValidateTransaction.kt
```diff
@@ -0,0 +1,18 @@
+package org.stellar.walletsdk.util
+
+import org.stellar.sdk.Server
+import org.stellar.sdk.Transaction
+
+fun validateTransaction(transaction: Transaction, server: Server) {
```
JakeUrban, line 6: Can we rename this `validateSufficientBalance()` or something similar? The name is too generic for what is does today.

JakeUrban (commented):
A few minor comments.

In general, I think we need to refactor the utils and bring them into the `Wallet` class. But that can be done in a separate PR.

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
quietbits: Not really... The `server` is defined for the whole class and it doesn't make sense to name it `defaultServer` in other places where it is used. This `serverInstance` is only used for tests and I couldn't come up with a better name. :thinking: 

image

src/main/kotlin/org/stellar/walletsdk/util/FetchAccountFromAddress.kt
quietbits: We'll create proper exceptions once the basic flow is working and we know exactly what kind of exceptions we need.

src/main/kotlin/org/stellar/walletsdk/Wallet.kt
JakeUrban: Oh hmm.. I would replace the `horizonUrl` parameter with a `server` parameter then, what way you provide a way to pass custom servers in tests but don't have a custom parameter on one of the wallet functions.

quietbits: `Wallet` class might change but I think it's too early to optimize it now. We don't have more significant pieces yet (create and recover accounts) that might drive some of the decisions.
For an average user, it might be easier (at least to me) to pass a `horizonUrl` rather than `server`. We can discuss this later with more people. :slightly_smiling_face:

PR #6, opened 2022-09-15
Update readme

quietbits:


README.md
```diff
@@ -139,8 +139,9 @@ val transaction = wallet.removeAccountSigner(sourceAddress, signerAddress)
 
 ```kotlin
 fun submitTransaction(
-    signedTransaction: Transaction
-): SubmitTransactionResponse
+    signedTransaction: Transaction,
+    serverInstance: Server = server
```
leighmcculloch, line 143: Is this `= server` a way to specify a default server? That's cool! Will we do same on the other functions that accept a server too?

leighmcculloch (approved):

piyalbasu (approved):

README.md
quietbits: Yup! `serverInstance` might not be the best name for it but it's OK for now. Right now it is used only for mocking server for tests. But we can add this to other methods as well if we find it valuable.

JakeUrban (approved):

PR #8, opened 2022-09-19
WalletSigner interface implementation POC

lijamie98:
This is a POC of the WalletSigner interface.

The naming maybe a little rough, so feel free to rename.


quietbits (approved):
I like this approach a lot. :rocket: I'll merge it in and do any renaming in my branch. Thanks, @lijamie98 !

PR #10, opened 2022-09-22
Handle async methods

quietbits:
Added suspend methods to make Android happy

### DIFF
--- a/src/test/kotlin/org/stellar/walletsdk/WalletTest.kt
+++ b/src/test/kotlin/org/stellar/walletsdk/WalletTest.kt
@@ -30,7 +30,7 @@ internal class WalletTest : SuspendTest() {
 
       // Public key
       assertEquals(56, publicKey.length)
-      assertTrue(publicKey.startsWith(""))
+      assertTrue(publicKey.startsWith("G"))
 
       // Secret key
       assertEquals(56, secretKey.length)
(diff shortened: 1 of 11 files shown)


JakeUrban (commented):
I haven't reviewed yet, but at a high level, is `suspend` & `runBlocking` andriods way of specifying sync vs. async code?

quietbits:
> I haven't reviewed yet, but at a high level, is `suspend` & `runBlocking` andriods way of specifying sync vs. async code?

Yes, `suspend` notes that the function is async. `runBlocking` is used only for testing. On Android, it's up to the user to decide which scope/thread to use when calling a `suspend` function.

```kotlin
screenScope.launch { // call suspend function here }
```

src/test/kotlin/org/stellar/walletsdk/WalletTest.kt
```diff
@@ -29,7 +30,7 @@ internal class WalletTest {
 
       // Public key
       assertEquals(56, publicKey.length)
-      assertTrue(publicKey.startsWith("G"))
+      assertTrue(publicKey.startsWith(""))
```
piyalbasu, line 33: Jw what the thought here is. Is this placeholder until you have a more inclusive test that includes muxed accts, for example?

piyalbasu (approved):

src/test/kotlin/org/stellar/walletsdk/WalletTest.kt
quietbits: That's my mistake, we should have "G" there. Thanks for catching it!

JakeUrban (approved):

PR #11, opened 2022-09-27
Recover account

quietbits:


src/main/kotlin/org/stellar/walletsdk/util/GsonUtils.kt
```diff
@@ -18,51 +12,5 @@ object GsonUtils {
 
   private fun builder(): GsonBuilder {
     return GsonBuilder()
-      .registerTypeAdapter(Instant::class.java, InstantConverter())
-      .registerTypeAdapter(Duration::class.java, DurationConverter())
-  }
-}
-
-class DurationConverter : JsonSerializer<Duration?>, JsonDeserializer<Duration?> {
```
quietbits, line 26: These methods were crashing the Android app. Removing them as we're not currently using them.

src/main/kotlin/org/stellar/walletsdk/util/CreateDecoratedSignature.kt
```diff
@@ -0,0 +1,26 @@
...
+  signatureBase64String: String,
+  base64Decoder: ((String) -> ByteArray)? = null
+): DecoratedSignature {
+  val signature = Signature()
+
+  if (base64Decoder == null) {
+    signature.signature = Base64.getDecoder().decode(signatureBase64String)
+  } else {
+    signature.signature = base64Decoder(signatureBase64String)
+  }
```
JakeUrban, line 19: Why would someone pass their own `base64Decoder`? It seems too low-level for the wallet SDK.

JakeUrban (commented):
Other than my one comment, I think we're missing some tests for this functionality right?

Also, I know I've mentioned this before, but I _really_ think we need to include an account as part of the `Wallet` object so we can move all these functions out of a `Utils` package and onto the `Wallet` object.

src/main/kotlin/org/stellar/walletsdk/util/CreateDecoratedSignature.kt
quietbits: `java.util.Base64` that we're using in Wallet SDK is supported only in Android API 23+. Anything older will need to use `android.util.Base64`. We did that in the [demo](https://github.com/quietbits/android-wallet-demo-test/blob/main/app/src/main/java/com/example/androidwalletdemo/view/RecoverAccount.kt#L45:L47) since we need to support Android API 21. I thought it was better to offer a fallback than ask users to build decorated signatures themselves. I'll add a note about it in the readme once I add this method, hopefully, that will make it clear enough.

quietbits:
> Other than my one comment, I think we're missing some tests for this functionality right?

@JakeUrban, yes, the tests are missing for methods that are making network requests. I haven't found a quick way to mock `okhttp` requests, so leaving that for later.

> Also, I know I've mentioned this before, but I _really_ think we need to include an account as part of the `Wallet` object so we can move all these functions out of a `Utils` package and onto the `Wallet` object.

Agree that we'll need to move some things around. It's hard to predict what we'll need. :sweat_smile: I'll put aside some time for cleanup once we have create/recover account in Wallet SDK.

src/main/kotlin/org/stellar/walletsdk/util/CreateDecoratedSignature.kt
JakeUrban: Is there a way to detect what android API we're using? Or can we try `java.util.Base64` and if that fails due to lack of support, try `andriod.util.Base64`?

quietbits: We can't use `android.util.Base64` in Kotlin Wallet SDK.

JakeUrban: Oh I see.

JakeUrban (approved):

lijamie98 (approved):
LGTM.

PR #15, opened 2022-10-19
Custom exceptions

quietbits:


src/main/kotlin/org/stellar/walletsdk/Exception.kt
```diff
@@ -0,0 +1,57 @@
...
+class InvalidMemoIdException : Exception("Memo ID must be a positive integer")
+
+class InvalidSponsorOperationTypeException(operationType: String, allowedOperations: List<String>) :
+  Exception(
+    "$operationType cannot be sponsored. Allowed operations are: ${allowedOperations
+      .joinToString(", ")}."
+  )
+
+class InvalidStartingBalanceException :
+  Exception("Starting balance must be at least 1 XLM for non-sponsored accounts")
```
piyalbasu, line 32: this might be too in the weeds for this use case, but is it worth considering subentries for minimum balance?

https://developers.stellar.org/docs/fundamentals-and-concepts/lumens#base-reserves

piyalbasu (approved):

src/main/kotlin/org/stellar/walletsdk/Exception.kt
quietbits: This exception is for an invalid minimum balance during account creation. We have a method that calculates reserved balance, which takes into account subentries. Is there a use case where we would need to calculate subentries for account creation (setting starting account balance)? Or maybe the exception name is not clear for the intended use case? :thinking:

piyalbasu: Oh got it. I didn't quite understand the context of this being during `fund` during Account Creation, but looking at where this is thrown, it makes more sense now. 👍

PR #16, opened 2022-10-20
KDoc documentation for classes and methods

quietbits:
First attempt to document classes and methods with KDoc.

### DIFF
--- a/src/main/kotlin/org/stellar/walletsdk/Auth.kt
+++ b/src/main/kotlin/org/stellar/walletsdk/Auth.kt
@@ -11,6 +11,21 @@ import org.stellar.sdk.Transaction
 import org.stellar.walletsdk.util.GsonUtils
 import org.stellar.walletsdk.util.OkHttpUtils
 
+/**
+ * Authenticate to an external server using
+ * [SEP-10](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md).
+ *
+ * @property accountAddress Stellar address of the account authenticating
+ * @property authEndpoint Authentication endpoint URL
+ * @property homeDomain Domain hosting stellar.toml (
+ * [SEP-1](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0001.md)) file
+ * containing `WEB_AUTH_ENDPOINT` URL and `SIGNING_KEY`
+ * @property memoId optional memo ID to distinguish the account
+ * @property clientDomain optional domain hosting stellar.toml file containing `SIGNING_KEY`
+ * @property networkPassPhrase Stellar network passphrase
+ * @property walletSigner interface to define wallet client and domain (if using `clientDomain`)
+ * signing methods
+ */
 class Auth(
   private val accountAddress: String,
   private val authEndpoint: String,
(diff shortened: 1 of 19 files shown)


src/main/kotlin/org/stellar/walletsdk/Auth.kt
```diff
@@ -11,6 +11,21 @@ import org.stellar.sdk.Transaction
 import org.stellar.walletsdk.util.GsonUtils
 import org.stellar.walletsdk.util.OkHttpUtils
 
+/**
+ * Authenticate to an external server using
+ * [SEP-10](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md).
+ *
+ * @property accountAddress Stellar address of the account authenticating
+ * @property authEndpoint Authentication endpoint URL
```
piyalbasu, line 19: It might be helpful to call this param `webAuthEndpoint` to align with the naming convention used in the SEP's so users who are learning can recognize what this maps to

Here's a good article that Leigh shared with me: https://martinfowler.com/bliki/UbiquitousLanguage.html

src/main/kotlin/org/stellar/walletsdk/util/CreateDecoratedSignature.kt
```diff
@@ -5,6 +5,17 @@ import org.stellar.sdk.KeyPair
 import org.stellar.sdk.xdr.DecoratedSignature
 import org.stellar.sdk.xdr.Signature
 
+/**
+ * Create decorated signature using account's Stellar address (public key) and signature string.
+ *
+ * @param publicKey Stellar address of the account to use in the signature
```
piyalbasu, line 11: Related to my comment above, I think we should call this param `accountAddress` (or something similar) since we refer to public keys as "addresses" in other instances in the SDK

src/main/kotlin/org/stellar/walletsdk/util/SetRecoveryMethods.kt
```diff
@@ -6,6 +6,26 @@ import kotlinx.coroutines.async
 import okhttp3.OkHttpClient
 import org.stellar.walletsdk.*
 
+/**
+ * Helper to register account with a recovery server using
+ * [SEP-30](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0030.md) and
+ * [SEP-10](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md).
+ *
+ * @param endpoint Recovery server's endpoint URL
```
piyalbasu, line 14: Call this `recoveryServerEndpoint` to be more descriptive?

src/main/kotlin/org/stellar/walletsdk/util/SetRecoveryMethods.kt
```diff
@@ -6,6 +6,26 @@ import kotlinx.coroutines.async
...
+ * @param authEndpoint Recovery server's auth endpoint URL of the recovery server to use with SEP-10
```
piyalbasu, line 15: Similar comment as above, `webAuthEndpoint`?

piyalbasu (approved):

src/main/kotlin/org/stellar/walletsdk/util/SetRecoveryMethods.kt
quietbits: All of those params are for the recovery server. Should we prefix all of them with `recoveryServer`? It seemed a bit redundant to repeat the "recovery server" in them. Maybe it would help with clarity? What do you think, @piyalbasu ?

piyalbasu: Oh gotcha. Yeah, let's not do that. Hmmmm, I was just trying to find a way to make `endpoint` more descriptive, but I guess the SEP doesn't really provide a better naming convention. Ignore my comment - the @param description should hopefully be enough

PR #18, opened 2022-10-26
SEP-24 interactive deposit and withdrawal

quietbits:
Deposit
https://user-images.githubusercontent.com/7346473/198392788-fe222eeb-07db-420a-b8c9-06aaf827a70f.mp4

Withdrawal
https://user-images.githubusercontent.com/7346473/198392842-6022f00c-b11e-473e-9f5b-b33a350b1e7e.mp4

Notes:
- The Android emulator might cause the UI issues we see in the anchor flow. I don't see those issues in the browser.
- The withdrawal process is incomplete (missing sending funds to the anchor step). The demo will be updated when the transaction status method is implemented.

### DIFF
--- a/src/main/kotlin/org/stellar/walletsdk/Anchor.kt
+++ b/src/main/kotlin/org/stellar/walletsdk/Anchor.kt
@@ -1,0 +1,166 @@
+package org.stellar.walletsdk
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
+import okhttp3.HttpUrl
+import okhttp3.HttpUrl.Companion.toHttpUrl
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import org.stellar.sdk.Network
+import org.stellar.sdk.Server
+import org.stellar.walletsdk.util.GsonUtils
+import org.stellar.walletsdk.util.OkHttpUtils
+import org.stellar.walletsdk.util.StellarToml
+
+//  TODO: move to Data
+data class AnchorServiceAsset(
+  val enabled: Boolean,
+  val min_amount: Double,
+  val max_amount: Double,
+  val fee_fixed: Double,
+  val fee_percent: Double
+)
+
+data class AnchorServiceFee(val enabled: Boolean)
+
+data class AnchorServiceFeatures(val account_creation: Boolean, val claimable_balances: Boolean)
+
+data class AnchorServiceInfo(
+  val deposit: Map<String, AnchorServiceAsset>,
+  val withdraw: Map<String, AnchorServiceAsset>,
+  val fee: AnchorServiceFee,
+  val features: AnchorServiceFeatures,
+)
+
+data class InteractiveResponse(
+  val id: String,
+  val url: String,
+  val type: String,
+)
+
+enum class StellarTomlFields(val text: String) {
+  SIGNING_KEY("SIGNING_KEY"),
+  TRANSFER_SERVER_SEP0024("TRANSFER_SERVER_SEP0024"),
+  WEB_AUTH_ENDPOINT("WEB_AUTH_ENDPOINT")
+}
+
+// TODO: document
+class Anchor(
+  private val server: Server,
+  private val network: Network,
+  private val httpClient: OkHttpClient = OkHttpClient()
+) {
+  suspend fun getInfo(assetIssuer: String): Map<String, Any> {
+    return CoroutineScope(Dispatchers.IO)
+      .async {
+        val toml = StellarToml(stellarAddress = assetIssuer, server, httpClient)
+
+        return@async toml.getToml()
+      }
+      .await()
+  }
+
+  suspend fun getServicesInfo(serviceUrl: String): AnchorServiceInfo {
+    try {
+      val url = serviceUrl.toHttpUrl()
+      val urlBuilder = HttpUrl.Builder().scheme(url.scheme).host(url.host).port(url.port)
+
+      url.pathSegments.forEach { urlBuilder.addPathSegment(it) }
+      urlBuilder.addPathSegment("info")
+
+      val infoUrl = urlBuilder.build().toString()
+
+      val request = Request.Builder().url(infoUrl).build()
+      val gson = GsonUtils.instance!!
+
+      return CoroutineScope(Dispatchers.IO)
+        .async {
+          httpClient.newCall(request).execute().use { response ->
+            if (!response.isSuccessful) throw NetworkRequestFailedException(response)
+
+            val infoResponse = response.body!!.charStream()
+            return@async gson.fromJson(infoResponse, AnchorServiceInfo::class.java)
+          }
+        }
+        .await()
+    } catch (e: Exception) {
+      throw Exception("Service URL is invalid", e)
+    }
+  }
+
+  suspend fun getInteractiveDeposit(
+    accountAddress: String,
+    assetCode: String,
+    assetIssuer: String,
+    clientDomain: String? = null,
+    memoId: String? = null,
+    extraFields: Map<String, Any>? = null,
+    walletSigner: WalletSigner,
+  ): InteractiveResponse {
+    return CoroutineScope(Dispatchers.IO)
+      .async {
+        val sep24RequiredFields =
+          listOf(
+            StellarTomlFields.SIGNING_KEY.text,
+            StellarTomlFields.TRANSFER_SERVER_SEP0024.text,
+            StellarTomlFields.WEB_AUTH_ENDPOINT.text
+          )
+        val toml = StellarToml(stellarAddress = assetIssuer, server, httpClient)
+        val tomlContent = toml.getToml()
+
+        // TODO: check toml for required fields
+        toml.hasFields(fields = sep24RequiredFields, tomlContent = tomlContent)
+
+        val transferServerEndpoint =
+          tomlContent[StellarTomlFields.TRANSFER_SERVER_SEP0024.text].toString()
+
+        // TODO: check the /info endpoint for the asset deposit
+        val serviceInfo = getServicesInfo(transferServerEndpoint)
+        val depositAsset =
+          serviceInfo.deposit[assetCode]
+            ?: throw Exception("Asset $assetCode is not accepted for deposits")
+
+        if (!depositAsset.enabled) {
+          throw Exception("Asset $assetCode is not enabled for deposits")
+        }
+
+        val homeDomain = toml.getHomeDomain()
+
+        // TODO: SEP-10 auth
+        val auth =
+          Auth(
+            accountAddress = accountAddress,
+            webAuthEndpoint = tomlContent[StellarTomlFields.WEB_AUTH_ENDPOINT.text].toString(),
+            homeDomain = homeDomain,
+            clientDomain = clientDomain,
+            memoId = memoId,
+            networkPassPhrase = network.networkPassphrase,
+            walletSigner = walletSigner
+          )
+        val authToken = auth.authenticate()
+
+        val requestParams = mutableMapOf<String, Any>()
+        requestParams["account"] = accountAddress
+        requestParams["asset_code"] = assetCode
+
+        if (extraFields != null) {
+          requestParams += extraFields
+        }
+
+        // TODO: get SEP-24 deposit link
+        val gson = GsonUtils.instance!!
+        val client = OkHttpClient()
+
+        val requestUrl = "$transferServerEndpoint/transactions/deposit/interactive"
+        val request = OkHttpUtils.buildJsonPostRequest(requestUrl, requestParams, authToken)
+
+        client.newCall(request).execute().use { response ->
+          if (!response.isSuccessful) throw NetworkRequestFailedException(response)
+
+          return@async gson.fromJson(response.body!!.charStream(), InteractiveResponse::class.java)
+        }
+      }
+      .await()
+  }
+}
--- a/src/main/kotlin/org/stellar/walletsdk/Anchor.kt
+++ b/src/main/kotlin/org/stellar/walletsdk/Anchor.kt
@@ -13,12 +13,25 @@ import org.stellar.walletsdk.util.GsonUtils
 import org.stellar.walletsdk.util.StellarToml
 import org.stellar.walletsdk.util.interactiveFlow
 
-// TODO: document
+/**
+ * Build on/off ramps with anchors.
+ *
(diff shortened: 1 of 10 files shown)


piyalbasu (approved):

src/main/kotlin/org/stellar/walletsdk/Anchor.kt
```diff
@@ -0,0 +1,172 @@
...
+   * @throws [AssetNotEnabledForDepositException] if asset is not enabled for deposits by the anchor
+   * @throws [NetworkRequestFailedException] if network request fails
+   */
+  suspend fun getInteractiveDeposit(
+    accountAddress: String,
+    assetCode: String,
+    assetIssuer: String,
+    memoId: String? = null,
+    clientDomain: String? = null,
+    extraFields: Map<String, Any>? = null,
```
piyalbasu, line 109: Jus curious because I was having trouble finding mention of it in the SEP: do we have any idea of what `extraFields` might be?

src/main/kotlin/org/stellar/walletsdk/util/StellarToml.kt
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * [Stellar info file](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0001.md)
+ * (also known as TOML file) methods.
+ *
+ * @property stellarAddress Stellar address whose TOML file to fetch
+ * @property server Horizon [Server] instance
+ * @property httpClient optional custom HTTP client, uses [OkHttpClient] by default
+ */
+class StellarToml(
+  private val stellarAddress: String,
```
JakeUrban, line 24: Why does the `StellarToml` class require a Stellar account as a parameter? Most anchors don't issue their own asset, so they don't have issuer accounts, and most wallets won't know the distribution accounts of the anchors they integrate with.

I think a more natural parameter would be `homeDomain`.

src/main/kotlin/org/stellar/walletsdk/util/StellarToml.kt
```diff
@@ -0,0 +1,123 @@
...
+   * Check if a list of fields exist in the TOML file.
+   *
+   * @param fields a list of fields to check
+   * @param tomlContent TOML file content
+   *
+   * @return `true` if all fields are found in the TOML file
+   *
+   * @throws [StellarTomlMissingFields] if there are any missing fields
+   */
+  fun hasFields(fields: List<String>, tomlContent: Map<String, Any>): Boolean {
```
JakeUrban, line 108: I don't think we should have this function, since its pretty simple to check if keys exist in a map.

Also, this function never returns `false`, and instead raises an exception if `true` cannot be returned. I don't think this is a good pattern.

src/main/kotlin/org/stellar/walletsdk/Anchor.kt
```diff
@@ -0,0 +1,172 @@
...
+            if (!response.isSuccessful) throw NetworkRequestFailedException(response)
+
+            val infoResponse = response.body!!.charStream()
+            return@async gson.fromJson(infoResponse, AnchorServiceInfo::class.java)
+          }
+        }
+        .await()
+    } catch (e: Exception) {
+      throw InvalidAnchorServiceUrl(e)
+    }
```
JakeUrban, line 81: We should really try and avoid wrapping large chunks of code in try blocks. Where is the line of code that would raise an exception resulting from an invalid URL?

src/main/kotlin/org/stellar/walletsdk/util/Sep24InteractiveFlow.kt
```diff
@@ -0,0 +1,127 @@
...
+  memoId: String? = null,
+  clientDomain: String? = null,
+  extraFields: Map<String, Any>? = null,
+  walletSigner: WalletSigner,
+  anchor: Anchor,
+  server: Server,
+  network: Network,
+  httpClient: OkHttpClient,
+): InteractiveFlowResponse {
+  return CoroutineScope(Dispatchers.IO)
```
JakeUrban, line 51: I'm assuming `CoroutineScope(Dispatchers.IO)` creates a thread that returns its results asynchronously right?

If so, shouldn't we only wrap the code that performs IO? For example, creating the `StellarToml` object doesn't need to be performed inside the async context. But if I have any incorrect assumptions please let me know!

src/main/kotlin/org/stellar/walletsdk/util/Sep24InteractiveFlow.kt
```diff
@@ -0,0 +1,127 @@
...
+        listOf(
+          StellarTomlFields.SIGNING_KEY.text,
+          StellarTomlFields.TRANSFER_SERVER_SEP0024.text,
+          StellarTomlFields.WEB_AUTH_ENDPOINT.text
+        )
+      val toml = StellarToml(stellarAddress = assetIssuer, server, httpClient)
+      val tomlContent = toml.getToml()
+
+      // Check toml for required SEP-24 fields
+      toml.hasFields(fields = sep24RequiredFields, tomlContent = tomlContent)
```
JakeUrban, line 63: Not all anchors issue their own assets (actually most don't), but this code is making the assumption that the asset issuer implements SEP-24. 

Circle, the USDC issuer, doesn't implement SEP-24. But MGI, which does implement SEP-24, doesn't issue their own asset and instead uses USDC.

src/main/kotlin/org/stellar/walletsdk/util/Sep24InteractiveFlow.kt
```diff
@@ -0,0 +1,127 @@
...
+          clientDomain = clientDomain,
+          memoId = memoId,
+          networkPassPhrase = network.networkPassphrase,
+          walletSigner = walletSigner,
+          httpClient = httpClient
+        )
+      val authToken = auth.authenticate()
+
+      val requestParams = mutableMapOf<String, Any>()
+      requestParams["account"] = accountAddress
```
JakeUrban, line 104: We're assuming here that `accountAddress` is both the account that is authenticating as well as the account that will be used as the source of funds. SEP-24 allows these accounts to be different though.

JakeUrban (changes requested):
Great to see SEP-24 support coming along! I think there are a few things that need adjustment though. Happy to discuss more.

src/main/kotlin/org/stellar/walletsdk/util/StellarToml.kt
```diff
@@ -0,0 +1,123 @@
...
+ * (also known as TOML file) methods.
+ *
+ * @property stellarAddress Stellar address whose TOML file to fetch
+ * @property server Horizon [Server] instance
+ * @property httpClient optional custom HTTP client, uses [OkHttpClient] by default
+ */
+class StellarToml(
+  private val stellarAddress: String,
+  private val server: Server,
+  private val httpClient: OkHttpClient = OkHttpClient()
```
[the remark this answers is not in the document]
quietbits: We need these for testing. We can't use this in tests if there is no way to pass mocked server or HTTP client.

src/main/kotlin/org/stellar/walletsdk/util/StellarToml.kt
quietbits: Thanks for catching this! I oversimplified it to make it easier on the UI (getting everything from the asset). I'll update to require `homeDomain` here.

src/main/kotlin/org/stellar/walletsdk/util/StellarToml.kt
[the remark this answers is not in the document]
JakeUrban: We should try to find a different solution for that, we can't let the need to test change our interface.

src/main/kotlin/org/stellar/walletsdk/util/StellarToml.kt
quietbits: I thought this helper might be useful for future SEP implementations where we need to make the same checks. The pattern is not great... I'm not sure how this should be handled in Kotlin properly. Throwing an exception seemed OK to me since it would give all the necessary details to fix it. I'll remove it and do this inline for now.

src/main/kotlin/org/stellar/walletsdk/Anchor.kt
quietbits: It's *only* 22 lines above. :wink: I'll update to move the exception closer to the source.

src/main/kotlin/org/stellar/walletsdk/util/Sep24InteractiveFlow.kt
quietbits: We're waiting for someone with more Kotlin experience to help us with coroutines. Hopefully, that will be soon. :crossed_fingers:

src/main/kotlin/org/stellar/walletsdk/Anchor.kt
```diff
@@ -0,0 +1,185 @@
...
+    val request = Request.Builder().url(infoUrl).build()
+    val gson = GsonUtils.instance!!
+
+    return CoroutineScope(Dispatchers.IO)
+      .async {
+        httpClient.newCall(request).execute().use { response ->
+          if (!response.isSuccessful) throw NetworkRequestFailedException(response)
+
+          val infoResponse = response.body!!.charStream()
+          return@async gson.fromJson(infoResponse, AnchorServiceInfo::class.java)
```
JakeUrban, line 81: Something to keep in mind here is that right now, you only have one `AnchorServiceInfo` class. But when the Wallet SDK supports SEP-6, 31, or 38, we'll have more.

That means that this function may need to attempt to deserialize the same response into a few different objects. Maybe that is acceptable, but it is inefficient. Ideally, we were able to know which object to use.

src/main/kotlin/org/stellar/walletsdk/util/Sep24InteractiveFlow.kt
```diff
@@ -0,0 +1,139 @@
...
+          accountAddress = accountAddress,
+          webAuthEndpoint = tomlContent[StellarTomlFields.WEB_AUTH_ENDPOINT.text].toString(),
+          homeDomain = homeDomain,
+          clientDomain = clientDomain,
+          memoId = memoId,
+          networkPassPhrase = network.networkPassphrase,
+          walletSigner = walletSigner,
+          httpClient = httpClient
+        )
+      val authToken = auth.authenticate()
```
JakeUrban, line 113: What if the wallet has already authenticated and has their own `authToken`? We should allow them to pass it instead of forcing them to go through SEP-10 again.

src/main/kotlin/org/stellar/walletsdk/util/StellarToml.kt
```diff
@@ -0,0 +1,73 @@
...
+    var scheme = "https"
+    var host = homeDomain
+
+    try {
+      val url = homeDomain.toHttpUrl()
+      scheme = url.scheme
+      host = url.host
+    } catch (e: Exception) {
+      // Use defaults
+    }
```
JakeUrban, line 69: Lets catch a specific exception. Also, maybe something like this?

```
    var scheme: String
    var host: String
    var url: ? // not sure what toHttpUrl() returns
    try {
      url = homeDomain.toHttpUrl()
    } catch (e: Exception) {
      url = null
    }
    if (url != null) {
      scheme = url.scheme
      host = url.host
    } else {
      scheme = "https"
      host = homedomain
    }
```

src/main/kotlin/org/stellar/walletsdk/util/Sep24InteractiveFlow.kt
```diff
@@ -0,0 +1,139 @@
...
+  return CoroutineScope(Dispatchers.IO)
+    .async {
+      val sep24RequiredFields =
+        listOf(
+          StellarTomlFields.SIGNING_KEY.text,
+          StellarTomlFields.TRANSFER_SERVER_SEP0024.text,
+          StellarTomlFields.WEB_AUTH_ENDPOINT.text
+        )
+      val toml = StellarToml(homeDomain, server, httpClient)
+      val tomlContent = toml.getToml()
```
JakeUrban, line 64: We should cache toml contents somehow, because it doesn't make sense to fetch the toml file every time we want to make a deposit request. Ideally we cache it on the `Anchor` class.

src/main/kotlin/org/stellar/walletsdk/util/Sep24InteractiveFlow.kt
JakeUrban: Or, we take the same approach as the toml contents and cache the auth token on the `Anchor` class.

src/main/kotlin/org/stellar/walletsdk/Anchor.kt
quietbits: Yes, we will need to use something generic here that can be passed when called.

src/main/kotlin/org/stellar/walletsdk/util/Sep24InteractiveFlow.kt
quietbits: :100: I thought the same once I started working on the status update polling. I'm not yet sure how to best handle that in Kotlin, but we will figure it out in the status update PR.

src/main/kotlin/org/stellar/walletsdk/util/Sep24InteractiveFlow.kt
quietbits: Yup! This is related to caching toml file. We will try to find a solution in the status update PR.

src/main/kotlin/org/stellar/walletsdk/util/StellarToml.kt
quietbits: :+1: There is a TODO on line 56 that would handle this better. I couldn't find a solid solution in Kotlin for this (in TypeScript it works a bit differently).

src/main/kotlin/org/stellar/walletsdk/util/Sep24InteractiveFlow.kt
JakeUrban: How does the transaction status polling logic relate to caching the toml or auth token?

JakeUrban (approved):

src/main/kotlin/org/stellar/walletsdk/Anchor.kt
Ifropc: @JakeUrban in future we may want to use polymorphic de-serialization, think it's a good TODO.

PR #19, opened 2022-11-03
Reorganize project structure

Ifropc:
1. Migrated gradle files to kotlin
2. Split project into multi-module (for future sub-modules such as examples)
3. Use libs.versions (as in java anchor sdk project) for store dependencies and versions
4. Upgrade gradlew 
5. Remove application plugin (not needed anymore)


settings.gradle.kts
```diff
@@ -0,0 +1,4 @@
+rootProject.name = "WalletSdk"
+
+/** SDK */
+include("core")
```
JakeUrban, line 4: I don't love the `core` name. Is there any issue with just naming it `wallet-sdk`? I don't think we should split wallet SDK functionality between subprojects, so that should be fine right?

JakeUrban (approved):
One question about the subproject name, otherwise looks good!

settings.gradle.kts
Ifropc: I will rename it to `wallet-sdk`. 
Also, for now the artifact for maven (to be used by external folks in the future) has name of
`org.stellar.wallet-sdk:core:0.1.0` 
I think I should rename it to
`org.stellar:wallet-sdk:0.1.0` as well?

PR #20, opened 2022-11-04
SEP-24 example

Ifropc:
- Added basic examples folder (so we can add more examples later)
- Example of (almost) full SEP-24 flow. 
- Didn't include sending transfer transaction for withdrawal, as it's not yet implemented. 
- Didn't include waiting for deposit transaction, also not implemented yet

### DIFF
--- a/examples/SEP-24/src/main/kotlin/org/stellar/walletsdk/Main.kt
+++ b/examples/SEP-24/src/main/kotlin/org/stellar/walletsdk/Main.kt
@@ -1,0 +1,59 @@
+package org.stellar.walletsdk
+
+import kotlinx.coroutines.runBlocking
+import org.stellar.sdk.KeyPair
+import org.stellar.sdk.Network
+import org.stellar.sdk.Server
+import org.stellar.sdk.Transaction
+
+private val myAddress = "GCNVZJLJ5XZZALLZBR7BGUMGMF2KVZTBIGTU2A5KOPS2PPCZS6PV4LAD"
+private val myKey = "SDYGC4TW5HHR5JA6CB2XLTTBF2DZRH2KDPBDPV3D5TXM6GF7FBPRZF3I"
+
+fun main() {
+  val wallet = Wallet()
+  val account = KeyPair.random()
+  val tx = runBlocking { wallet.fund(myAddress, account.publicKeyString, "10") }
+
+  tx.sign(KeyPair.fromSecretSeed(myKey))
+
+  val res = runBlocking { wallet.submitTransaction(tx) }
+
+  val anchor = Anchor(Server("https://horizon-testnet.stellar.org"), Network.TESTNET)
+
+  val info = runBlocking {
+    anchor.getInfo("https://testanchor.stellar.org/.well-known/stellar.toml")
+  }
+
+  val servicesInfo = runBlocking { anchor.getServicesInfo("https://testanchor.stellar.org/sep24") }
+
+  val resp = runBlocking {
+    anchor.getInteractiveWithdrawal(
+      account.publicKeyString,
+      homeDomain = "testanchor.stellar.org",
+      assetCode = "SRT",
+      walletSigner = WalletSignerImpl(account)
+    )
+  }
+
+  print(resp)
+}
+
+val KeyPair.publicKeyString: String
+  get() = this.accountId
+
+val KeyPair.secretKey: String
+  get() = this.secretSeed.concatToString()
+
+class WalletSignerImpl(private val keyPair: KeyPair) : WalletSigner {
+  override fun signWithClientAccount(txn: Transaction): Transaction {
+    txn.sign(keyPair)
+    return txn
+  }
+
+  override fun signWithDomainAccount(
+    transactionString: String,
+    networkPassPhrase: String
+  ): Transaction {
+    TODO("Not yet implemented")
+  }
+}
--- a/examples/SEP-24/src/main/kotlin/org/stellar/walletsdk/Main.kt
+++ b/examples/SEP-24/src/main/kotlin/org/stellar/walletsdk/Main.kt
@@ -68,7 +68,8 @@ suspend fun main() {
   // val transaction = GET TRANSFER_SERVER_SEP0024/transaction
 
   println("Waiting for tokens...")
-  // Wait for token to appear on user account
+  // Optional step: wait for token to appear on user account
+  // TODO: replace with waiting for transaction
   while (
     server
       .accounts()
--- a/examples/SEP-24/src/main/kotlin/org/stellar/walletsdk/Main.kt
+++ b/examples/SEP-24/src/main/kotlin/org/stellar/walletsdk/Main.kt
@@ -11,13 +11,14 @@ import org.stellar.sdk.Transaction
 // Setup main account that will fund new (user) accounts. You can get new key pair and fill it with
 // testnet tokens at
 // https://laboratory.stellar.org/#account-creator?network=test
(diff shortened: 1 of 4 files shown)


examples/SEP-24/src/main/kotlin/org/stellar/walletsdk/Main.kt
```diff
@@ -0,0 +1,122 @@
...
+// Setup main account that will fund new (user) accounts. You can get new key pair and fill it with
+// testnet tokens at
+// https://laboratory.stellar.org/#account-creator?network=test
+private val myAddress = "GCNVZJLJ5XZZALLZBR7BGUMGMF2KVZTBIGTU2A5KOPS2PPCZS6PV4LAD"
+private val myKey = "SDYGC4TW5HHR5JA6CB2XLTTBF2DZRH2KDPBDPV3D5TXM6GF7FBPRZF3I"
+
+suspend fun main() {
+  // Create instance of server that allows to connect to Horizon
+  val server = Server("https://horizon-testnet.stellar.org")
+  val wallet = Wallet()
```
quietbits, line 20: Just a heads up, the API for `Wallet` changes slightly in [this PR](https://github.com/stellar/kotlin-wallet-sdk/pull/21).

examples/SEP-24/src/main/kotlin/org/stellar/walletsdk/Main.kt
```diff
@@ -0,0 +1,122 @@
...
+  // Generate new (user) account and fund it with 10 XLM from main account
+  val account = KeyPair.random()
```
quietbits, line 22: We could use `Wallet.create()` method to generate a new keypair.

examples/SEP-24/src/main/kotlin/org/stellar/walletsdk/Main.kt
```diff
@@ -0,0 +1,122 @@
...
+
+  // Sign with your main account's key and send transaction to the network
+  println("Registering new account")
+  tx.sign(KeyPair.fromSecretSeed(myKey))
+  assert(wallet.submitTransaction(tx))
+
+  val anchor = Anchor(server, Network.TESTNET)
+
+  // Get info from the anchor server
+  val info = anchor.getInfo("https://testanchor.stellar.org/.well-known/stellar.toml")
```
quietbits, line 33: Passing only the home domain `testanchor.stellar.org` will work. We're appending `.well-known/stellar.toml` internally. Should we update description to make it more clear?

quietbits (approved):
Looks good! Glad that it works. :sweat_smile:

examples/SEP-24/src/main/kotlin/org/stellar/walletsdk/Main.kt
```diff
@@ -0,0 +1,122 @@
...
+  while (
+    server
+      .accounts()
+      .account(account.publicKeyString)
+      .balances
+      .filter { it.assetCode.isPresent && it.assetCode.get() == "SRT" }
+      .any { it.balance.toBigDecimal() <= BigDecimal.ZERO }
+  ) {
+    delay(5.seconds)
+  }
```
JakeUrban, line 82: I don't quite understand this code. It looks like the `.any` line is checking for amounts less than 0.

examples/SEP-24/src/main/kotlin/org/stellar/walletsdk/Main.kt
```diff
@@ -0,0 +1,122 @@
...
+  override fun signWithDomainAccount(
+    transactionString: String,
+    networkPassPhrase: String
+  ): Transaction {
+    val txn =
+      Transaction.fromEnvelopeXdr(transactionString, Network(networkPassPhrase)) as Transaction
+    txn.sign(keyPair)
+    return txn
+  }
+}
```
JakeUrban, line 122: @Ifropc you actually don't need to implement `signWithDomainAccount()`. I wonder how we can make this more clear @quietbits ?

examples/SEP-24/src/main/kotlin/org/stellar/walletsdk/Main.kt
```diff
@@ -0,0 +1,122 @@
...
+  // Generate new (user) account and fund it with 10 XLM from main account
+  val account = KeyPair.random()
+  val tx = wallet.fund(myAddress, account.publicKeyString, "10")
+
+  // Sign with your main account's key and send transaction to the network
+  println("Registering new account")
+  tx.sign(KeyPair.fromSecretSeed(myKey))
+  assert(wallet.submitTransaction(tx))
+
+  val anchor = Anchor(server, Network.TESTNET)
```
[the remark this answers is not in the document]
quietbits: I agree. An empty `Anchor` class is not very helpful. This would simplify inner methods.

examples/SEP-24/src/main/kotlin/org/stellar/walletsdk/Main.kt
quietbits: Both override methods need to be there, but `signWithDomainAccount` can be empty. Is there a magic way we could handle it better in Kotlin? We can update the description with a note about what is optional.

examples/SEP-24/src/main/kotlin/org/stellar/walletsdk/Main.kt
Ifropc: It creates AccountKeyPair and it's not usable with java sdk as for now (i.e. can't sign transactions with it)

examples/SEP-24/src/main/kotlin/org/stellar/walletsdk/Main.kt
Ifropc: I think so. 
IMO it's a good idea to pass home domain to Anchor so we can just call anchor.info

examples/SEP-24/src/main/kotlin/org/stellar/walletsdk/Main.kt
```diff
@@ -0,0 +1,122 @@
...
+import org.stellar.sdk.KeyPair
+import org.stellar.sdk.Network
+import org.stellar.sdk.Server
+import org.stellar.sdk.Transaction
+
+// Setup main account that will fund new (user) accounts. You can get new key pair and fill it with
+// testnet tokens at
+// https://laboratory.stellar.org/#account-creator?network=test
+private val myAddress = "GCNVZJLJ5XZZALLZBR7BGUMGMF2KVZTBIGTU2A5KOPS2PPCZS6PV4LAD"
+private val myKey = "SDYGC4TW5HHR5JA6CB2XLTTBF2DZRH2KDPBDPV3D5TXM6GF7FBPRZF3I"
```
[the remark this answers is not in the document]
Ifropc: It was intended for anybody who is running this example to set it manually (as it's usually easier this way). I added functionality with getting key from env variable.

examples/SEP-24/src/main/kotlin/org/stellar/walletsdk/Main.kt
Ifropc: Yes, it waits until account has a balance of SRT token > 0
It's not a good solution in general, in future it will be replaced with waiting for transfer transaction (there is a TODO there)

examples/SEP-24/src/main/kotlin/org/stellar/walletsdk/Main.kt
Ifropc: Yes, current interface is extensive. We should refactor it later.

PR #29, opened 2022-11-21
Exceptions refactoring

Ifropc:
I tried grouping all exceptions in multiple categories. For now, there is a root `WalletException` (all exceptions that are thrown in the code extend it)
Then, there are multiple groups of exceptions:
1. `ValidationException` -> invalid user input 
2. `AnchorAssetException` (all exceptions related to asset errors on the anchor) -> to be extended to `AnchorException` (?) 
3. `RecoveryException` -> errors returned from recovery server
4. `StellarException` -> errors returned from Horizon
5. `InvalidResponseException` -> malformed response from the server (mainly, missing Json fields)
6. `NetworkRequestFailedException` -> non-200 code from servers (can extend later if needed) 
7. `Uncategorized` (not belong to any other groups, inherit `WalletException` directly). Currently only `NetworkRequestFailedException` and `StellarTomlMissingFields`

I think a good strategy is logically group several exception types together (e.g. `AnchorAssetException`) so that they can have a generic handler.


quietbits (approved):
Nice!

PR #31, opened 2022-11-24
Setup maven publish

Ifropc:
Configured publishing to maven central

### DIFF
--- a/wallet-sdk/build.gradle.kts
+++ b/wallet-sdk/build.gradle.kts
@@ -53,7 +53,7 @@ publishing {
 
   publications {
     create<MavenPublication>("mavenJava") {
-      groupId = "org.stellar.wallet-sdk"
+      groupId = "org.stellar"
       artifactId = "wallet-sdk"
 
       from(components["java"])
--- a/wallet-sdk/build.gradle.kts
+++ b/wallet-sdk/build.gradle.kts
@@ -3,6 +3,7 @@
 plugins {
   `maven-publish`
   alias(libs.plugins.dokka)
+  signing
 }
 
 dependencies {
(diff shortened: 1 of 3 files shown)


wallet-sdk/build.gradle.kts
```diff
@@ -31,13 +32,71 @@ val javadocJar =
...
+        username = System.getenv("OSSRH_USER") ?: return@credentials
+        password = System.getenv("OSSRH_PASSWORD") ?: return@credentials
+      }
+    }
+  }
+
   publications {
     create<MavenPublication>("mavenJava") {
-      groupId = "org.stellar"
+      groupId = "org.stellar.wallet-sdk"
```
lijamie98, line 56: The `groupId` should be `org.stellar` because the `artifactId` is `wallet-sdk`

wallet-sdk/build.gradle.kts
```diff
@@ -31,13 +32,71 @@ val javadocJar =
...
       from(components["java"])
+      artifact(javadocJar)
+      artifact(sourcesJar)
+
+      pom {
+        name.set("Stellar Wallet SDK")
+        description.set("Kotlin Stellar Wallet SDK")
+        licenses {
+          license {
+            name.set("The Apache License, Version 2.0")
```
lijamie98, line 68: I think we should use Apache 2.0. But you might want to check with Legal.

wallet-sdk/build.gradle.kts
Ifropc: It's been like this initially, but I thought that if in the future we introduce more packages, it's better if group id doesn't change. I can set it back to org.stellar

quietbits: I used Anchor Platform as an example for `build.gradle` configuration. `group` is set to `org.stellar.anchor-sdk` [here](https://github.com/stellar/java-stellar-anchor-sdk/blob/develop/build.gradle.kts#L124). Is that the same one we're using here for `groupId`? I'll let you decide what is the correct value here. :slightly_smiling_face:

Ifropc: I will keep it `org.stellar:wallet-sdk:0.1.0` for now
We can later transition to `org.stellar.wallet-sdk:<artifact>:<version>` later

lijamie98 (approved):
LGTM. It will be nice to double check with @quietbits

quietbits (approved):

PR #33, opened 2022-11-25
Restructure auth and anchor

Ifropc:
Restructure auth and anchor. Contains #32 changes.


wallet-sdk/src/main/kotlin/org/stellar/walletsdk/Auth.kt
```diff
@@ -13,26 +13,20 @@ import org.stellar.walletsdk.util.OkHttpUtils
...
-  private val accountAddress: String,
   private val webAuthEndpoint: String,
   private val homeDomain: String,
-  private val memoId: String? = null,
-  private val clientDomain: String? = null,
-  private val networkPassPhrase: String = Network.TESTNET.toString(),
   private val walletSigner: WalletSigner,
+  private val network: Network = Network.TESTNET,
   private val httpClient: OkHttpClient = OkHttpClient()
 ) {
```
quietbits, line 31: What is the benefit of splitting these params between the class and inner methods? Is it to make the params list shorter?

To make the class more generic, we could move `walletSigner` to `authenticate` method.

wallet-sdk/src/main/kotlin/org/stellar/walletsdk/anchor/Sep24InteractiveFlow.kt
```diff
@@ -0,0 +1,165 @@
+package org.stellar.walletsdk.util
```
quietbits, line 1: Should we move away from using SEPs in file/method names? Interactive flow == SEP 24

wallet-sdk/src/test/kotlin/org/stellar/walletsdk/AnchorTest.kt
```diff
@@ -77,18 +78,18 @@ internal class AnchorTest {
...
+        anchor
+          .interactive()
+          .deposit(
             accountAddress = ADDRESS_ACTIVE,
-            toml = toml,
-            walletSigner = InProcessWalletSigner()
+            fundsAccountAddress = ADDRESS_ACTIVE_TWO,
+            assetCode = ANCHOR_ASSET_CODE,
+            authToken = authToken,
           )
```
quietbits, line 92: Looking at how this is used, it might be convenient to have params passed to `interactive(...)`, then `deposit` or `withdraw` can be called without any params (they are the same for both of them). Thoughts?

wallet-sdk/src/main/kotlin/org/stellar/walletsdk/anchor/Sep24InteractiveFlow.kt
Ifropc: Yes, I agree. I just forgot to rename the file.

wallet-sdk/src/main/kotlin/org/stellar/walletsdk/Auth.kt
Ifropc: Yes, one of them is to make a list shorter.
But in general, the idea is that `Auth` object should be re-used. For example, I see that the flow should be
1. Create `Auth` class (one per server -> already pre-defined endpoints, etc)
2. Call `authenticate()` function for every user (i.e. once per account)

In general, class params -> generic configuration (per whole app lifecycle, per app, per server, per user, etc)
Method params -> less generic (one level down)

I didn't move `walletSigner`, because I think it should be static per server (unless `signWithDomainAccount` is called? In that case you are right, it can be moved to function parameters)

wallet-sdk/src/test/kotlin/org/stellar/walletsdk/AnchorTest.kt
Ifropc: Yes, but user can won't be able to re-use Interactive object in that case

quietbits (approved):

PR #47, opened 2022-12-07
Restructure wallet

Ifropc:
This PR includes some big restructure. It splits existing logic into separate classes, accessible via root configurable `Wallet` class
List below is classes after restructure, old name (if changed), description, example of usages

Wallet (new) -> root class. Is used to spawn all other classes. Similar to Server from Java SDK
```kotlin
wallet = Wallet(StellarConfiguration.Testnet, ApplicationConfiguration(WalletSignerImpl()) // TODO: create default implementation of WalletSigner
```
StellarConfiguration -> configures all Stellar network parameters (such as connection to Horizon, which network app is operating
ApplicationConfiguration -> more higher-level configuration, defines defaults to be used across the app.
Config -> internal class, contains stellar and app configuration. Is propagated to all classes, so they know about shared configuration
```kotlin
class Anchor
internal constructor(
  private val cfg: Config
  // <...>
)
  private val server = cfg.stellar.server
  private val network = cfg.stellar.network
```
WalletSigner -> class that is used to sign transaction by target address for auth request. Default implementation to be added (will use address private key to sign. Assumes account is single-signature, user can define default WalletSigner to be used per-app, as well as per-auth request) 
 
Anchor -> represents connection to one anchor server
```kotlin
val anchor = wallet.anchor(homeUrl)
```
Recovery -> represents SEP-30 features (work with recovery servers. Currently servers are NOT bound per-object. In the future, they should be)
```kotlin
val recovery = wallet.recovery()
```
Auth -> represents SEP-10 authentication by server. Used internally by Recovery (in the future we may want to lift it up in the flow, to allow re-using tokens). Provided by Anchor
```kotlin
val auth = anchor.auth().authenticate(address)
```

Stellar -> represents connection to Horizon server. 
```kotlin
val stellar = wallet.stellar()
val server = stellar.server // underlying Horizon server connection
stellar.submitTransaction(tx) // only method available 
```

AccountService (Previously some functions from Wallet) (?) -> works with stellar accounts. 
```kotlin
val account = stellar.account()
val keyPair = account.create()
val info = account.getInfo(keyPair.address)
val history = account.getHistory(keyPair.address)
```

TransactionBuilder (Previously most functions from Wallet) -> builds stellar transactions
```kotlin
val txBuilder = stellar.transaction()
```

Example: add trustline
```kotlin
val addTrustline = txBuilder.addAssetSupport(keypair.address, asset)
addTrustline.sign(keypair)
stellar.submitTransaction(addTrustline)
```


JakeUrban:
Gleb, this is great. I think it takes a big step in the right direction making the wallet SDK easy to understand.

I don't see the definitions of `ApplicationConfiguration` or `StellarConfiguration` though, am I missing something?

quietbits (approved):
:rocket:

Ifropc:
> Gleb, this is great. I think it takes a big step in the right direction making the wallet SDK easy to understand.
> 
> I don't see the definitions of `ApplicationConfiguration` or `StellarConfiguration` though, am I missing something?

It's currently in Wallet file (because of big diff it didn't render)
https://github.com/stellar/kotlin-wallet-sdk/blob/9c2361bd8464bcfd0ca80d67c3d232fbf8affbac/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/Wallet.kt#L44

PR #55, opened 2022-12-22
Documentation: Stellar transaction builder + account service

quietbits:


docs/WalletGuide.md
```diff
@@ -40,3 +48,115 @@ val walletCustom = Wallet(StellarConfiguration.Testnet, ApplicationConfiguration
...
+import org.stellar.walletsdk.asset.IssuedAssetId
+import org.stellar.walletsdk.horizon.sign
+
+suspend fun main() { 
+-->
+<!--- SUFFIX .*stellar.*
+}    
+-->
+
+## Build on Stellar
```
Ifropc, line 63: Can you split account and transaction builder into 2 files?
Also, formatting with 2 spaces in the beginning would also be nice

docs/WalletGuide.md
```diff
@@ -40,3 +48,115 @@ val walletCustom = Wallet(StellarConfiguration.Testnet, ApplicationConfiguration
 <!--- INCLUDE
 val wallet = Wallet(StellarConfiguration.Testnet)
 -->
+
+<!--- INCLUDE .*stellar.*
+import org.stellar.walletsdk.*
```
Ifropc, line 53: FYI: this import shouldn't be necessary (already defined above)

quietbits: Somehow, it's not adding this import from the global INCLUDE.

Ifropc (approved):

PR #59, opened 2023-01-09
Cleanup info

Ifropc:
Now info is cached and not needed to be re-fetched every time


quietbits (approved):
:rocket: :tada:

Ifropc:
WAL-600

PR #69, opened 2023-02-09
Basic TS example

quietbits:


Ifropc:
Can you add package.json and stuff like that so it's possibly to actually run the example?

PR #78, opened 2023-03-09
[0.6.0][WAL-573] Migrate to Ktor client

Ifropc:


quietbits (approved):

docs/WalletGuide.md
```diff
@@ -490,7 +541,7 @@ suspend fun main() {
 Build on and off ramps with anchors for deposits and withdrawals.
 
 ```kotlin
-val anchor = wallet.anchor("testanchor.stellar.org")
+val anchor = wallet.anchor(Url("https://testanchor.stellar.org"))
```
JakeUrban, line 544: This seems like something we can do inside the `anchor()` function. Why do we need to have the user wrap the URL string in the `Url` class?

docs/WalletGuide.md
```diff
@@ -49,7 +54,52 @@ There is one more available configuration for wallet that allows to configure in
...
+  StellarConfiguration.Testnet,
+  ApplicationConfiguration { defaultRequest { url { protocol = URLProtocol.HTTP } } }
+)
+```
+
+### Configuring client
+Wallet SDK uses [ktor client](https://ktor.io/docs/getting-started-ktor-client.html) for all network requests 
+(excluding Horizon, where SDK http client is used). Currently, okhttp engine is configured to be used with the client.  
+Ktor has a reach client configuration, you can read about it more [here](https://ktor.io/docs/create-client.html#configure-client).
+For example, client can be globally configured:
```
JakeUrban, line 67: I think we're missing 'the' in a few places. I also added a few phrasing changes

```suggestion
The Wallet SDK uses the [ktor client](https://ktor.io/docs/getting-started-ktor-client.html) for all network requests 
(excluding Horizon, where the Stellar SDK's http client is used). Currently, the okhttp engine is configured to be used with the client.  
You can read more about how to configure the ktor client [here](https://ktor.io/docs/create-client.html#configure-client).
For example, the client can be globally configured:
```

docs/WalletGuide.md
```diff
@@ -49,7 +54,52 @@ There is one more available configuration for wallet that allows to configure in
...
+        }
+      }
+    )
+  )
+```
+This code will set connect timeout to 10 seconds via 
+[okhttp configuration](https://ktor.io/docs/http-client-engines.html#okhttp) 
+and install [retry plugin](https://ktor.io/docs/client-retry.html)  
+Some services also allows client configuration. For example, to change connect timeout when connecting to recovery 
+servers:
```
JakeUrban, line 87:
```suggestion
This code will set the connect timeout to 10 seconds via the
[okhttp configuration](https://ktor.io/docs/http-client-engines.html#okhttp) 
and also installs the [retry plugin](https://ktor.io/docs/client-retry.html).
You can also specify client configuration for specific Wallet SDK classes. For example, to change connect timeout when connecting to recovery 
servers:
```

docs/WalletGuide.md
```diff
@@ -49,7 +54,52 @@ There is one more available configuration for wallet that allows to configure in
...
+val recoveryCustomClient =
+  walletCustomClient.recovery {
+    engine { this.config { this.connectTimeout(Duration.ofSeconds(30)) } }
+  }
+```
+
+### Closing resources
+After wallet class is no longer used, it's necessary to close all clients used by it. While in some applications it may 
+not be required (e.g. wallet lives for the whole lifetime of the app), in other cases it can be required.  
+If your wallet class is short-lived it's recommended to close client resources using close function:
```
JakeUrban, line 98:
```suggestion
After the wallet class is no longer used, it's necessary to close all clients used by it. While in some applications it may 
not be required (e.g. the wallet lives for the whole lifetime of the app), in other cases it can be required.  
If your wallet class is short-lived, it's recommended to close client resources using close function:
```

wallet-sdk/src/main/kotlin/org/stellar/walletsdk/Wallet.kt
```diff
@@ -96,22 +124,49 @@ internal data class Config(val stellar: StellarConfiguration, val app: Applicati
...
+      }
+    }
+  )
+
+  val defaultClient =
+    HttpClient(OkHttp) {
+      install(ContentNegotiation) { json(defaultJson) }
+      defaultRequest { url { protocol = URLProtocol.HTTPS } }
+      defaultClientConfig()
+    }
```
JakeUrban, line 163: It also might make sense to set reasonable defaults for things like the connect timeout. Check out the Python Stellar SDK [client](https://stellar-sdk.readthedocs.io/en/stable/_modules/stellar_sdk/client/requests_client.html#RequestsClient) and its defaults.

Ifropc: We use the defaults from okhttp client

docs/WalletGuide.md
Ifropc: I'm still thinking on how to organize it better. 
The thing is that we need to unify links. The easiest way to do it is to have Url type. Because when it's string, it's unclear if it's just domain or if it's full path .
The other way that I used in previous commit was
```kotlin
anchor({ host = "testanchor.stellar.org" })
```
Not sure if it's better or not

Ifropc: I guess the best option would be to accept string and try to understand if it's full path or just a host. 
Though, if user passes something unexpected, there could be some obscure errors

JakeUrban: Stellar uses the concept of home domains in a lot of different places. Stellar account's have a `home_domain` attribute, which points too the domain hosting your SEP-1 toml file. Home domains are also passed in SEP-10 for client attribution. I think we should try to continue using domains in the Wallet SDK to keep this pattern since it's already familiar.

Ifropc: I will do it like this. It'll be allowed to pass either home domain or full url

Ifropc: It accepts string again (in either format)

JakeUrban (approved):

PR #84, opened 2023-04-06
[Wal-587] [WAL-666] Recovery re-work + documentation

Ifropc:


docs/WalletGuide.md
```diff
@@ -699,3 +707,224 @@ suspend fun main() {
...
+        mapOf(first to identities, second to identities),
+        SignerWeight(10, 5),
+        sponsor
+      )
+    )
+```
+With given parameters, this function will:
+1. Create `account` on the Stellar network (if it's not created yet).
+2. `sponsor` will be used to sponsor all operations. You may not sponsor creating recoverable wallet, but in that case `account` must already be created.
+3. `deviceKey` will be used as a primary account key. Please note that master key belonging to `account` will be locked. `deviceKey` should be used as a primary signer instead.
```
JakeUrban, line 798: Isn't this pretty opinionated? Couldn't an account use its master key as the device key instead of generating a new one? I understand this is how vibrant did it but I never understood why.

docs/WalletGuide.md
```diff
@@ -699,3 +707,224 @@ suspend fun main() {
...
+4. Set all account thresholds to 10. You can read more about threshold in the [documentation](https://developers.stellar.org/docs/encyclopedia/signatures-multisig#thresholds)
+5. Use identities that were defined earlier on both servers. (That means, both server will accept SEP-10 authentication via `recoveryKey` as an auth method)
+6. Set device key weight to 10, and recovery server weight to 5. Given account thresholds, both servers must be used to recover account, as transaction signed by one will only have weight of 5, which is not sufficient to change account key.
+
+Finally, sign and submit transaction to the network
+```kotlin
+  val tx = recoverableWallet.transaction.sign(account).sign(sponsor)
```
JakeUrban, line 805: Q: could we change the `sign()` function to take in a variable number of arguments?
```suggestion
  val tx = recoverableWallet.transaction.sign(account, sponsnor)
```

docs/WalletGuide.md
JakeUrban: What happens if I pass the account's master secret key in as the device key?

docs/WalletGuide.md
```diff
@@ -699,3 +707,224 @@ suspend fun main() {
...
+-->
+
+<!--- SUFFIX 
+}
+-->
+
+You can fetch account info from one or more servers. To do so, first we need to authenticate,
+```kotlin
+  val auth1 = recovery.sep10Auth(first).authenticate(recoveryKey)
+  val auth2 = recovery.sep10Auth(second).authenticate(recoveryKey)
```
JakeUrban, line 839: I don't understand where `first` and `second` come from.

docs/WalletGuide.md
```diff
@@ -699,3 +707,224 @@ suspend fun main() {
...
+```
+We need to know signers that will be used to sign the transaction. You can get them either on wallet creation step (`recoverySigners` variable), or via fetching account info from recovery serves. 
+
+Next, create a new device key and retrieve a signed key replacement transaction from recovery servers:
+```kotlin
+  val newKey = wallet.stellar().account().createKeyPair()
+
+  val serverAuth = mapOf(
+    first to RecoveryServerSigning(recoverySigners[0], auth1),
+    second to RecoveryServerSigning(recoverySigners[1], auth2)
```
JakeUrban, line 900: Is this `first to ` kotlin syntax? How does it work?

docs/WalletGuide.md
```diff
@@ -699,3 +707,224 @@ suspend fun main() {
...
+    recovery.replaceDeviceKey(
+      account,
+      newKey,
+      serverAuth
+    )
+```
+Calling with function will create a transaction that locks previous device key and replaces it with a new key (having the same weight as an old one). 
+Lost device key is deduced automatically. Signer will be considered a device key, if one of these conditions matches:
+1. It's the only signer that's not in `serverAuth`
+2. All signers in `serverAuth` has the same weight, and potential signer is the only one with a different weight.
```
JakeUrban, line 913: What if the old device key doesn't match the criteria here? Can't we specify the old device public key as a parameter to avoid ambiguity?

JakeUrban (approved):
Great documentation 👏  I left some comments & questions that might lead to adjustments, but after discussion there may not be anything blocking.

docs/WalletGuide.md
Ifropc: I decided to leave it like this for now. It's pretty common approach, and the good thing about it is that it's well paired with a method that creates transaction for signer recovery.
I was thinking about making it optional, but it will complicate implementation. We may add it in the future vesions

Ifropc: > What happens if I pass the account's master secret key in as the device key?

It will throw an error, but probably not very informative. I will look into it.

docs/WalletGuide.md
Ifropc: I think it's defined earlier in the document

docs/WalletGuide.md
Ifropc: Yes

docs/WalletGuide.md
Ifropc: Yes, it's alias to
Pair(first, RecoveryServerSigning)

quietbits (approved):

docs/WalletGuide.md
Ifropc: If we change it to vararg, it will accept 0 signers as an argument (won't be checked in compile time)
I don't really like it, so will keep current implementation

PR #92, opened 2023-06-08
[WAL-868] Improve transaction submission

Ifropc:
Add handler for https://github.com/stellar/java-stellar-sdk/blob/master/src/main/java/org/stellar/sdk/Server.java#L282


JakeUrban (approved):
LGTM!

PR #93, opened 2023-06-12
[ANCHOR-305] Add support for native assets

philipliu:
This adds XLM support for the SEP-24 client.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -16,7 +16,7 @@ val jvmVersion = JavaVersion.VERSION_1_8
 
 allprojects {
   group = "org.stellar.wallet-sdk"
-  version = "0.8.0-snapshot"
+  version = "0.8.0-SNAPSHOT"
 }
 
 subprojects {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -16,7 +16,7 @@ val jvmVersion = JavaVersion.VERSION_1_8
 
 allprojects {
   group = "org.stellar.wallet-sdk"
-  version = "0.7.0"
+  version = "0.8.0-snapshot"
 }
 
 subprojects {
--- a/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/anchor/Interactive.kt
+++ b/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/anchor/Interactive.kt
@@ -43,10 +43,9 @@ internal constructor(
     fundsAccountAddress: String? = null,
   ): InteractiveFlowResponse {
     return flow(accountAddress, assetId, authToken, extraFields, fundsAccountAddress, "withdraw") {
-      if (assetId is IssuedAssetId) {
-        it.withdraw[assetId.code]
-      } else {
-        it.withdraw[NativeAssetId.id]
+      when (assetId) {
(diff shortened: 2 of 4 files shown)


wallet-sdk/src/main/kotlin/org/stellar/walletsdk/anchor/Interactive.kt
```diff
@@ -35,13 +37,17 @@ internal constructor(
    */
   suspend fun withdraw(
     accountAddress: String,
-    assetId: IssuedAssetId,
+    assetId: StellarAssetId,
```
Ifropc, line 40: We have 3 types of asset ids: stellar, issued and fiat.
Can we either add interface that only stellar and issued implements, or replace "if" when "when" (for exhaustive pattern matching)

wallet-sdk/src/main/kotlin/org/stellar/walletsdk/anchor/Interactive.kt
```diff
@@ -35,13 +37,17 @@ internal constructor(
...
     authToken: AuthToken,
     extraFields: Map<String, String>? = null,
     fundsAccountAddress: String? = null,
   ): InteractiveFlowResponse {
     return flow(accountAddress, assetId, authToken, extraFields, fundsAccountAddress, "withdraw") {
-      it.withdraw[assetId.code]
+      if (assetId is IssuedAssetId) {
```
Ifropc, line 46: nit: can replace lambda type with:
`assetGet: (AnchorServiceInfo, String) -> AnchorServiceAsset?`
second argument is either assetId.code or NativeAssetId.id

wallet-sdk/src/main/kotlin/org/stellar/walletsdk/anchor/Interactive.kt
philipliu: I can make the pattern match more exhaustive. I didn't think it was necessary because `StellarAssetId` is either `NativeAssetId` or `IssuedAssetId`, but it's better to be safe if we ever change the SEP-38 format.

Ifropc: I didn't see you were using StellarAssetId.
Then it's LGTM.
Still I prefer to use `when` over `if` (in case we add more types)

wallet-sdk/src/main/kotlin/org/stellar/walletsdk/anchor/Interactive.kt
philipliu: I'll leave it as is to minimize the changes

build.gradle.kts
```diff
@@ -16,7 +16,7 @@ val jvmVersion = JavaVersion.VERSION_1_8
 
 allprojects {
   group = "org.stellar.wallet-sdk"
-  version = "0.7.0"
+  version = "0.8.0-snapshot"
```
Ifropc, line 19: Sorry for the wrong version I gave you, can you capitalize it?
0.8.0-SNAPSHOT

Ifropc (approved):

PR #95, opened 2023-06-14
[WAL-881] Add default implementation of domain http signer

Ifropc:
With this change, wallets would be able to implement client_domain signing as simply as creating new instance of the implementation
```kotlin
anchor
    .auth()
    .authenticate(
      wallet.stellar().account().createKeyPair(),
      WalletSigner.DomainSigner("https://demo-wallet-server.stellar.org/sign"),
      clientDomain = "demo-wallet-server.stellar.org"
    )
```

### DIFF
--- a/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/auth/WalletSigner.kt
+++ b/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/auth/WalletSigner.kt
@@ -1,5 +1,14 @@
 package org.stellar.walletsdk.auth
 
+import io.ktor.client.*
+import io.ktor.client.call.*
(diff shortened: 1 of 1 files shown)


JakeUrban (approved):
Does this support the wallet adding some level of authentication to these requests? That would be critical here. Maybe we should even try to enforce it.

Ifropc:
Yes, they need to write `requestTransformer` lambda. Right now optional, but I can make it mandatory.
They write something like this:
```
WalletSigner.DomainSigner("https://demo-wallet-server.stellar.org/sign") { bearerAuth("token")}
```

Ifropc:
Update: transformer is mandatory now

PR #102, opened 2023-06-21
[WAL-916] Watcher implementation

Ifropc:
Add watcher class that allows to watch transaction status updates.
Example of usage:
```kotlin
suspend fun watchTransaction() {
  val watcher = anchor.watcher()
  val result = watcher.watchOneTransaction(getAuthToken(), "transaction id")

  do {
    val event = result.channel.receive()
    when (event) {
      is StatusChange ->
        println("Status changed to ${event.status}. Transaction: ${event.transaction}")
      is ChannelClosed -> println("Channel closed. Job is done")
      is ExceptionHandlerExit -> println("Exception handler exited the job")
    }
  } while (event !is ChannelClosed)
}
```

### DIFF
--- a/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/anchor/Anchor.kt
+++ b/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/anchor/Anchor.kt
@@ -93,7 +93,7 @@ internal constructor(
   /**
    * Creates new transaction watcher
    *
-   * @param pullDelay pull interval in which requests to the Anchor are being made.
+   * @param pollDelay poll interval in which requests to the Anchor are being made.
    * @param channelSize size of the Coroutine [Channel]. See
    *   [channel documentation](https://kotlinlang.org/docs/coroutines-and-channels.html#channels)
    *   for more info about channel size configuration. Be default, unlimited channel is created.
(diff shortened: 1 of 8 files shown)


wallet-sdk/src/main/kotlin/org/stellar/walletsdk/anchor/Anchor.kt
```diff
@@ -82,6 +91,25 @@ internal constructor(
     return Interactive(this, httpClient)
   }
 
+  /**
+   * Creates new transaction watcher
+   *
+   * @param pullDelay pull interval in which requests to the Anchor are being made.
```
JakeUrban, line 97: I think you mean `poll`, not `pull`, right?

JakeUrban (approved):
LGTM, but I wonder if it makes sense to consolidate the polling logic in both `watchOneTransaction()` and `watchAsset()`. Seems like the logic is the same other than the request made and the criteria to evaluate if it should exit.

wallet-sdk/src/main/kotlin/org/stellar/walletsdk/anchor/Anchor.kt
Ifropc: Yep

Ifropc:
>I wonder if it makes sense to consolidate the polling logic
I tried to refactor this place, but there is a variable re-assignment inside "logical" part, which complicates refactoring too much (so I decided to leave it as is)

PR #107, opened 2023-07-06
[WAL-944] Reorganization

Ifropc:
All functions with SEP-24 functionality has been moved to `Interactive` class (responsible for SEP-24)
Also, added SEP12 alias for Customer class/function

### DIFF
--- a/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/anchor/Anchor.kt
+++ b/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/anchor/Anchor.kt
@@ -32,7 +32,7 @@ internal constructor(
   private val baseUrl: Url,
   private val httpClient: HttpClient
 ) {
-  private val infoHolder = InfoHolder(cfg.stellar.network, baseUrl, httpClient)
+  internal val infoHolder = InfoHolder(cfg.stellar.network, baseUrl, httpClient)
 
   /**
    * Get anchor information from a TOML file.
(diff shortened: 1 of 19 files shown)


wallet-sdk/src/main/kotlin/org/stellar/walletsdk/anchor/Anchor.kt
```diff
@@ -266,13 +173,19 @@ internal constructor(
   }
 }
 
+suspend fun Anchor.sep12(authToken: AuthToken): Sep12 {
+  return this.customer(authToken)
+}
+
 fun Anchor.sep24(): Sep24 {
   return this.interactive()
 }
```
JakeUrban, line 182: Why do we encourage the use of `anchor.interactive()` vs `anchor.sep24()`? The latter seems to be more straightforward, and makes it easy to add more SEPs without defining some one-word alias to it.

Ifropc: It's just how it's been historically. I can rename it to sep24 and keep current alias there

JakeUrban (approved):

wallet-sdk/src/main/kotlin/org/stellar/walletsdk/anchor/Anchor.kt
CassioMG: That's a good point, I'm also in favor of using `sep24()`. Could we please also rename the `Interactive` class to `Sep24` instead? So it's clear that it's the class that interfaces with all Sep-24 related stuff - what do you think?

Ifropc: Yep that's the idea 👍🏼

CassioMG (approved):

PR #134, opened 2024-06-12
Add `transaction_id` to SEP-12 requests

philipliu:
This is to support the latest version of the SEP-12 protocol. Tests have not been added because the test anchor does not support the `transaction_id` fields.

### DIFF
--- a/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/customer/Sep12.kt
+++ b/wallet-sdk/src/main/kotlin/org/stellar/walletsdk/customer/Sep12.kt
@@ -25,6 +25,7 @@ internal constructor(
    * memo is present in the decoded SEP-10 JWT's sub value, it must match this parameter value.
    * @param type (optional) the type of action the customer is being KYCd for.
    * @param lang (optional) Defaults to en. Language code specified using ISO 639-1. Human-readable
+   * @param transactionId (optional) the ID of the transaction that the customer is being KYC'ed for.
    * descriptions, choices, and messages should be in this language.
    * @return a customer data object
    */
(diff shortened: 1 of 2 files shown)


Ifropc (approved):

wallet-sdk/src/main/kotlin/org/stellar/walletsdk/customer/Sep12.kt
```diff
@@ -69,19 +73,23 @@ internal constructor(
...
     type: String? = null,
+    transactionId: String? = null,
   ): AddCustomerResponse {
     val customer: MutableMap<String, String> = mutableMapOf()
 
     populateMap(type, customer, memo, sep9Info)
 
     val urlBuilder = URLBuilder(baseUrl)
     urlBuilder.appendPathSegments("customer")
+    urlBuilder.addParameter("transaction_id", transactionId)
```
CassioMG, line 92: @philipliu @Ifropc shouldn't the `transaction_id` value go in the request `body` (aka _populateMap_) instead of a url parameter for this `PUT` request?

wallet-sdk/src/main/kotlin/org/stellar/walletsdk/customer/Sep12.kt
```diff
@@ -116,6 +127,7 @@ internal constructor(
...
```
CassioMG, line 130: @philipliu @Ifropc same question here: shouldn't the transaction_id value go in the request body (aka populateMap) instead of a url parameter for this PUT request?

wallet-sdk/src/main/kotlin/org/stellar/walletsdk/customer/Sep12.kt
philipliu: Thanks for catching this. I'll publish a PR to fix this.

PR #156, opened 2024-11-14
Bump version to 2.0.0

philipliu:


Ifropc (approved):

build.gradle.kts
```diff
@@ -12,7 +12,7 @@ val jvmVersion = JavaVersion.VERSION_11
 
 allprojects {
   group = "org.stellar.wallet-sdk"
-  version = "1.7.2"
+  version = "1.8.0"
```
lijamie98, line 15: Should this be patch or minor?

philipliu: I think we want to start lining up the release numbers with the typescript-wallet-sdk https://github.com/stellar/typescript-wallet-sdk/releases/tag/v1.8.0.

Ifropc: I don't necessarily think we want to align versions with the wallet SDK. We can either make new wallet SDK version 2.0 or 1.8 but I vote for 1.8: we did some small breaking previously in minor versions, and I think it should be fine, given the scope of usage.

lijamie98: If it breaks the backward compatibility, we should increment the `major` version according to [semver](https://semver.org/).

lijamie98: This ticket [ANCHOR-865](https://stellarorg.atlassian.net/browse/ANCHOR-867) is designed to prevent the dependency conflicts.

philipliu: I've bumped the version to 2.0

lijamie98 (approved):
LGTM

PR #157, opened 2024-11-14
[ANCHOR-867] Create Uber JAR

lijamie98:
1. Create Uber JAR in the build.
2. Publish Uber JAR to Maven


Ifropc:
I think it'd be great to add GH actions for automatic deployment (it's not set up yet)

